PR #6, opened 2022-04-21
switch to abc.ABC-based protocols (GDEV-775)

KerstenBreuer:
- use abc.ABC instead of typing.Protocol as base classes for protocols and ports
- migrate ascii-validation into protocols
- refactor utils module to only include functionality used within the hexkit package


hexkit/protocols/eventsub.py
```diff
@@ -41,15 +42,17 @@ class MalformedPayloadError(RuntimeError):
...
+    @abstractmethod
+    def consume(self, *, payload: JsonObject, type_: Ascii, topic: Ascii) -> None:
         """Receive an event of interest and process it according to its type.
 
         Args:
             payload (JsonObject): The data/payload to send with the event.
             type_ (str): The type of the event.
             topic (str): Name of the topic to publish the event to.
         """
+        check_ascii([type_, topic])
```
Cito, line 57:
```suggestion
        check_ascii(type_, topic)
```
See suggestion for changing this in utils.

hexkit/utils.py
```diff
@@ -15,47 +15,18 @@
...
+    """Thrown when non-ASCII string was unexpectedly provided"""
 
-TEST_FILE_PATHS = [
-    TEST_FILE_DIR / filename
-    for filename in os.listdir(TEST_FILE_DIR)
-    if filename.startswith("test_") and filename.endswith(".yaml")
-]
+    def __init__(self, str_value: str):
+        """Prepare custom message."""
+        super().__init__(f"Non-ASCII string provided: {str_value}")
```
Cito, line 24:
```suggestion
        super().__init__(f"Non-ASCII string provided: {str_value!r}")
```
Better to use the repr() of the string in the message, particularly when it's not ASCII.

hexkit/utils.py
```diff
@@ -15,47 +15,18 @@
...
-    signal.alarm(timeout_after)
-
-    # execute the function
-    result = func(*func_args_, **func_kwargs_)
-
-    # disable the timer
-    signal.alarm(0)
-
-    return result
+def check_ascii(str_values: list[str]):
```
Cito, line 27:
```suggestion
def check_ascii(*str_values: str):
```
This makes usage a bit simpler, particularly when you want to check a single string.

tests/unit/test_utils.py
```diff
@@ -0,0 +1,37 @@
...
+    "str_values, exception",
+    [
+        (["valid", "also_valid_123-?$3&"], None),
+        (["valid", "invälid"], NonAsciiStrError),
+    ],
+)
+def test_check_ascii(str_values: list[str], exception: Optional[Exception]):
+    """Test the check_ascii function"""
+    with pytest.raises(exception) if exception else nullcontext():
+        check_ascii(str_values)
```
Cito, line 37:
```suggestion
        check_ascii(*str_values)
```
Adapt to suggested change of signature.

Cito (approved):
Well done. Just some minor suggestions below.

hexkit/utils.py
KerstenBreuer: Great comment, thanks

hexkit/utils.py
KerstenBreuer: makes total sense, thanks

KerstenBreuer:
@Cito, thanks a lot for the thorough review, as always. Highly appreciate it.

PR #8, opened 2022-04-22
Properly close Kafka Producers and Consumers

Cito:
On my machine, the integration tests crash at the end with the following error message:

```
tests/integration/test_akafka.py::test_kafka_event_subscriber
/home/vscode/.local/lib/python3.9/site-packages/_pytest/unraisableexception.py:78: PytestUnraisableExceptionWarning: Exception ignored in: <function ConsumerCoordinator._del_ at ...>
```

This seems to happen because the garbage collector kicks in and the Kafka consumer is not closed.

So we should make sure that consumers and producers are always closed in the implementation fo the providers and when used directly in the integration tests.

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -14,7 +14,7 @@ services:
     privileged: true
 
     volumes:
-      - ..:/workspace:cached
+      - .:/workspace:cached
 
     # Overrides default command so things don't shut down after the process ends.
     command: sleep infinity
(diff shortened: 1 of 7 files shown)


Cito:
> Generally, I find it a bit strange that the tests run through on my machine and in the CI. Where you running inside the devcontainer? Would like to understand what caused this different behaviors.

Yes, I ran the tests in the devcontainer. It's probably a race condition. The error happens in the `__del__` method of the ConsumerCoordinator class, which is always a troublemaker and can produce nondeterministic behavior.

PR #11, opened 2022-05-16
utilities for async dependency injection (GDEV-845)

KerstenBreuer:
```
- Added a class that extends the `dependency_injector`'s Resource provider
   to support classes with a `construct` method that returns an async
   context manager
- Added a class adds an async context manager interface to the
   `DeclarativeContainer` of the `dependency_injector` framework to handle
   the setup and teardown of resources
- Added a helper function that automatically chooses the correct constructor
   for a given class
````

### DIFF
--- a/hexkit/inject.py
+++ b/hexkit/inject.py
@@ -32,6 +32,14 @@ import dependency_injector.providers
 
 from hexkit.custom_types import ContextConstructable
 
+__all__ = [
+    "get_constructor",
+    "ContainerBase",
+    "NotConstructableError",
+    "ContextConstructor",
+    "AsyncInitShutdownError",
+]
+
 
 class NotConstructableError(TypeError):
     """Thrown when a ContextConstructable expected but not obtained."""
--- a/hexkit/inject.py
+++ b/hexkit/inject.py
@@ -1,0 +1,80 @@
+# Copyright 2021 - 2022 Universität Tübingen, DKFZ and EMBL
+# for the German Human Genome-Phenome Archive (GHGA)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""Utilities for dependency injection based on the dependency_injector framework."""
+
+from typing import Any, AsyncContextManager, AsyncIterator, Callable, Optional
+
+import dependency_injector.providers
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: tests/unit/test_inject.py [hunk 1]
```diff
@@ -0,0 +1,131 @@
...
+    constructable: ContextConstructable, exception: Optional[Exception]
+):
+    """Test whether init and shutdown correctly works with a context constructor."""
+
+    foo = "bar"
+
+    test = ContextConstructor(constructable, foo)
+
+    with pytest.raises(exception) if exception else nullcontext():  # type: ignore
+        await test()
```
Cito, line 108:
```suggestion
        resource = await test.async_()
        assert isinstance(resource, constructable)
```

hexkit/inject.py
```diff
@@ -0,0 +1,183 @@
...
+
+import inspect
+from typing import Any, AsyncContextManager, AsyncIterator, Callable, Optional, TypeVar
+
+import dependency_injector.containers
+import dependency_injector.providers
+
+from hexkit.custom_types import ContextConstructable
+
+
```
Cito, line 35: You should add the things that his module exports for usage here:

```python
__all__ = [
    'get_constructor', 'ContainerBase',
    'NotConstructableError', 'AsyncInitShutdownError'
]
```

hexkit/inject.py
```diff
@@ -0,0 +1,183 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""
+Utilities for dependency injection based on the dependency_injector framework.
+
```
Cito, line 19: You should state the main purpose of the module here, something like:

"This module contains a helper function that automatically selects the suitable "Provider" class for the dependency_injector framework and makes the "Container" class usable as async context manager."

Cito (approved):
Looks good, I think that should do it for now.

I think it's important for hexkit to have good examples and docs, so that people don't need to read the source and understand all aspects of the dependency_injector framework.

hexkit/inject.py
```diff
@@ -0,0 +1,183 @@
...
+    async context manager.
+    """
+
+    if not (
+        hasattr(constructable, "construct")
+        # Using isinstance together with Callable should be supported,
+        # see: https://peps.python.org/pep-0484/#callable
+        # Thus ignoring mypy error:
+        and isinstance(constructable.construct, Callable)  # type: ignore
+    ):
```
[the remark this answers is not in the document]
KerstenBreuer: Nice

hexkit/inject.py
KerstenBreuer: Thanks, good point

KerstenBreuer: Done

hexkit/inject.py
KerstenBreuer: Done.

KerstenBreuer:
Thanks a lot @Cito for the nice review.

PR #12, opened 2022-06-02
made protocols and providers async (GDEV-716)

KerstenBreuer:
Description:
```
* Made eventpub protocol + Kafka provider async

* tested ascii validation of protocols separately

* Made eventsub protocol + Kafka provider asyn

* update github action deps

* adapt example
```

### DIFF
--- a/hexkit/protocols/eventpub.py
+++ b/hexkit/protocols/eventpub.py
@@ -45,7 +45,7 @@ class EventPublisherProtocol(ABC):
     async def _publish_validated(
         self, *, payload: JsonObject, type_: Ascii, key: Ascii, topic: Ascii
     ) -> None:
-        """Publish an event that has already been validated.
+        """Publish an event with already validated topic and type.
 
         Args:
             payload (JSON): The payload to ship with the event.
--- a/hexkit/protocols/eventpub.py
+++ b/hexkit/protocols/eventpub.py
@@ -26,7 +26,7 @@ class EventPublisherProtocol(ABC):
     """A protocol for publishing events to an event broker."""
 
     @abstractmethod
-    def publish(
+    async def publish(
         self, *, payload: JsonObject, type_: Ascii, key: Ascii, topic: Ascii
     ) -> None:
         """Publish an event.
--- a/hexkit/protocols/eventpub.py
+++ b/hexkit/protocols/eventpub.py
@@ -25,7 +25,6 @@ from hexkit.utils import check_ascii
 class EventPublisherProtocol(ABC):
     """A protocol for publishing events to an event broker."""
 
-    @abstractmethod
     async def publish(
         self, *, payload: JsonObject, type_: Ascii, key: Ascii, topic: Ascii
     ) -> None:
@@ -38,4 +37,20 @@ class EventPublisherProtocol(ABC):
             topic (str): The event type. ASCII characters only.
         """
         check_ascii(type_, key, topic)
+        await self._publish_validated(
+            payload=payload, type_=type_, key=key, topic=topic
+        )
+
+    @abstractmethod
+    async def _publish_validated(
+        self, *, payload: JsonObject, type_: Ascii, key: Ascii, topic: Ascii
+    ) -> None:
+        """Publish an event that has already been validated.
+
+        Args:
+            payload (JSON): The payload to ship with the event.
+            type_ (str): The event type. ASCII characters only.
+            key (str): The event type. ASCII characters only.
+            topic (str): The event type. ASCII characters only.
+        """
         ...
--- a/hexkit/protocols/eventsub.py
+++ b/hexkit/protocols/eventsub.py
@@ -46,7 +46,7 @@ class EventSubscriberProtocol(ABC):
     types_of_interest: list[Ascii]
 
     @abstractmethod
-    def consume(self, *, payload: JsonObject, type_: Ascii, topic: Ascii) -> None:
+    async def consume(self, *, payload: JsonObject, type_: Ascii, topic: Ascii) -> None:
         """Receive an event of interest and process it according to its type.
 
         Args:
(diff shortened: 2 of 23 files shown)


tests/unit/test_eventpub.py
```diff
@@ -0,0 +1,76 @@
...
+)
+async def test_ascii_val(type_, key, topic, exception):
+    """Tests the ASCII validation logic included in the EventPublisherProtocol."""
+    payload = {"test_content": "Hello World"}
+
+    # create event publisher:
+    event_publisher = FakePublisher()
+
+    # publish event using the provider:
+    with (pytest.raises(exception) if exception else nullcontext()):  # type: ignore
```
Cito, line 70: No outer parens needed here.

tests/unit/test_eventsub.py
```diff
@@ -0,0 +1,67 @@
...
+)
+async def test_ascii_val(type_, topic, exception):
+    """Tests the ASCII validation logic included in the EventSubscriberProtocol."""
+    payload = {"test_content": "Hello World"}
+
+    # create event publisher:
+    event_submitter = FakeSubscriber()
+
+    # publish event using the provider:
+    with (pytest.raises(exception) if exception else nullcontext()):  # type: ignore
```
Cito, line 62: No outer braces needed

tests/unit/test_in_mem_pub.py
```diff
@@ -16,44 +16,22 @@
...
-            "test_ßtopic",  # non ascii
-            NonAsciiStrError,
-        ),
-    ],
-)
-def test_in_mem_publisher(type_, key, topic, exception):
+
+
+@pytest.mark.asyncio
+async def test_in_mem_publisher():
```
Cito, line 29: See same comment for KafkaPublisher. The error tests may still make sense to check that super() is used by the InMemPublisher.

Cito (approved):
Nice. Added some minor remarks, but feel free to ignore them.

setup.cfg
```diff
@@ -49,7 +49,7 @@ python_requires = >= 3.9
 
 [options.extras_require]
 akafka =
-    kafka-python>=2.0.2
+    aiokafka>=0.7.2
```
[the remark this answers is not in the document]
KerstenBreuer: Good idea.

tests/unit/test_akafka.py
```diff
@@ -16,67 +16,38 @@
...
-)
-def test_kafka_event_publisher(
-    type_: str,
-    key: str,
-    topic: str,
-    expected_headers: list[tuple[str, bytes]],
-    exception: Optional[type[Exception]],
-):
+@pytest.mark.asyncio
+async def test_kafka_event_publisher():
```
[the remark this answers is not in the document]
KerstenBreuer: This is a good point. I was thinking a bit more and probably it's either way not a good design to rely on the implementer (in case of the publishing protocol that is us, but in case of the subscription protocol that is whomever is using our library) to call super().

Maybe it is cleaner to have the validation logic in a non-abstract method which is calling another private method. This private method is marked as abstract and must be implemented by the developer that subclasses from that abstract class.  
Like this we would nicely separate the concerns of users and implementors in two separate methods.

Please have a look at this commit, as an example: https://github.com/ghga-de/hexkit/pull/12/commits/d599600b66ff7838ed85da4e231cc99af3a9fe0e

tests/unit/test_akafka.py
```diff
@@ -86,33 +57,25 @@ def test_kafka_event_publisher(
...
             key=key,
-            headers=expected_headers,
+            topic=topic,
         )
-        producer.flush.assert_called_once()
-    # check if producer was closed correctly:
-    producer.close.assert_called_once()
+
+    # check if producer was correctly used:
+    producer.start.assert_awaited_once()
```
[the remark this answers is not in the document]
KerstenBreuer: Yes, in deed, that would be better.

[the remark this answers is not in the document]
KerstenBreuer: ✅

tests/unit/test_eventpub.py
KerstenBreuer: Right

KerstenBreuer: ✅

tests/unit/test_eventsub.py
KerstenBreuer: ✅

tests/unit/test_akafka.py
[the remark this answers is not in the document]
Cito: Yes, that's a nice solution. +1

hexkit/protocols/eventsub.py
```diff
@@ -55,4 +54,18 @@ def consume(self, *, payload: JsonObject, type_: Ascii, topic: Ascii) -> None:
...
         """
         check_ascii(type_, topic)
+        await self._consume_validated(payload=payload, type_=type_, topic=topic)
+
+    @abstractmethod
+    async def _consume_validated(
+        self, *, payload: JsonObject, type_: Ascii, topic: Ascii
+    ) -> None:
+        """
+        Receive a pre-validated event of interest and process it according to its type.
```
Cito, line 64: This description is a bit misleading, you might think that the event itself is or was validated.

"Receive and process an event with already validated topic and type."?

hexkit/protocols/eventpub.py
```diff
@@ -38,4 +37,20 @@ def publish(
             topic (str): The event type. ASCII characters only.
         """
         check_ascii(type_, key, topic)
+        await self._publish_validated(
```
Cito, line 40: This separation makes sense, let's do it.

Cito (approved):
Looks good to me. I would rephrase one docstring, see below.

hexkit/protocols/eventsub.py
KerstenBreuer: Agreed.

KerstenBreuer: ✅

hexkit/protocols/eventpub.py
KerstenBreuer: Great, thanks.

PR #22, opened 2022-08-22
adding transaction management to DAO (GDEV-1151)

KerstenBreuer:
For squash and merge:

```
The DAOs returned by the DaoFactoryProtocol now have
a `with_transaction` method that returns an async context
manager for managing transactions.

Thereby, upon __aenter__ it returns a transactionally scoped DAO.
This recursive structure supports having nested transactions.
Moreover, a provider might implement the DAO operations
performed with and without transactions in a different way
(which is for instance useful for MongoDB, which operates
much faster without than with transactions).

Also updated the find API by splitting it into two separate 
methods find_all and find_one. The first for getting all hits,
the second one for getting a single hit.
Option to return the newest and oldest hit where removed
(there is no easy and generic way to achieve that in 
MongoDB).

Bump version to 0.3.0.
```


hexkit/protocols/dao.py
```diff
@@ -124,56 +122,47 @@ async def delete(self, *, id_: str) -> None:
...
+            Returns a hit in the form of the respective DTO model or None if no hit
+            was found.
 
         Raises:
-            NoHitsFoundError:
-                Raised when no hits where found when used in "newest", "oldest", or
-                "single" mode. When using the "all" mode, zero hits will not cause an
-                exception but simply result in an empty list beeing returned.
             MultpleHitsFoundError:
                 Raised when obtaining more than one hit when using the "single" mode.
```
Cito, line 148:
```suggestion
                Raised when obtaining more than one hit.
```

hexkit/protocols/dao.py
```diff
@@ -184,6 +173,21 @@ class DaoSurrogateId(DaoCommons[Dto], typing.Protocol[Dto, DtoCreation_contra]):
      creation of new resources, is needed.
```
Cito, line 173: "is needed" -> i**f** needed?

Cito (approved):
LGTM. We can still revise the API when implementing the Mongo provider.

hexkit/protocols/dao.py
KerstenBreuer: Don't know what was written there :D deleted it

PR #28, opened 2022-08-26
configurable ID generation in DaoFactoryProtocol (GDEV-1204)

KerstenBreuer:
For squash and merge:

```
The `get_dao` method of the DaoFactoryProtocol now takes an
AsycGenerator that is used for generating new resource IDs.
```


Cito (changes requested):
I like the idea of using a generator instead of a simple function, because it allows to use something like a simple counter.

Currently the integration tests are failing - the reason is that the default value for id_generator is a global object which is a generator which can be used only once. Better to use None as default value which would indicate to use (a fresh instance of) the default uuid4 generator, or to use a generator function as default value, instead of a generator, i.e. a callable that returns a generator.

Cito (approved):
This looks good now. For the default id generator it does not matter if it is reused.

PR #29, opened 2022-08-31
standardize configuration of providers (GDEV-1206)

KerstenBreuer:
For squash and merge:
```
All providers now take a pydantic model as
config argument. Also added a Configurator that is specific for handling
configuration based on pydantic for dependency injection.
```


Cito (approved):
Looks good.

PR #36, opened 2022-10-13
async non-context constructable (GDEV-1357)

KerstenBreuer:
Description:
```
Sometimes a component needs async initialization/construction logic
(that cannot be handeled in a sync __init__ method).
So far, hexkit only supported this via a special classmethod that returns
an async context manager for setup or teardown.
However, often teardown is not required.
Thus hexkit now also excepts plain async constuctor methods
without returning an async context manager.

Bump version to 0.5.0.

```


hexkit/custom_types.py
```diff
@@ -31,13 +31,35 @@
...
+# A AsyncConstructable is a class with a async classmethod `construct` that is used when
+# asynchronous constuction/instantiation logic is needed (which cannot be handeled in
+# a synchronous __init__ method).
+# With the current typing features of Python, it seems not possible to correctly type
+# a class with that signature.
+# E.g. one could define a typing.Protocol subclass with following method:
+# ```
+# class AsyncConstructable(Protocol):
+#     @classmethod
+#     def construct(cls, *args: Any, **kwargs: Any): ...
```
Cito, line 43:
```suggestion
#     async def construct(cls, *args: Any, **kwargs: Any): ...
```

hexkit/custom_types.py
```diff
@@ -31,13 +31,35 @@
...
+# ```
+# However, this is incompatible with implementations that don't explicitly use `*args`
+# and `*kwargs`, e.g. the following method does not comply with the above function stub:
+# ```
+# class SomeImplementation:
+#     @classmethod
+#     def construct(cls, foo: str): ...
```
Cito, line 50:
```suggestion
# class SomeImplementation(AsyncConstructable):
#     @classmethod
#     async def construct(cls, foo: str): ...
```

hexkit/custom_types.py
```diff
@@ -31,13 +31,35 @@
...
+# ```
+# Thus using a type alias for now:
+AsyncConstructable = Any
```
Cito, line 53: This should work and is a little bit better:

```python
class AsyncConstructable(Protocol):
    """A class that can be asynchronously constructed.

    An AsyncConstructable is a class with an async classmethod `construct`
    that is used when asynchronous constuction/instantiation logic is needed
    (which cannot be handeled in a synchronous __init__ method).
    """

    construct: Callable[..., Awaitable["AsyncConstructable"]]
```

hexkit/custom_types.py
```diff
@@ -51,4 +73,4 @@
 #     def construct(cls, foo: str): ...
 # ```
 # Thus using a type alias for now:
-ContextConstructable = Any
+AsyncContextConstructable = Any
```
Cito, line 76: That's indeed hard to type properly because the contextlib really doesn't make it easy. This is the best I could come up with: 

```python
class AsyncContextConstructable(Protocol):
    """A class that can be asynchronously constructed with a context manager.

    An AsyncContextConstructable is a class with a classmethod `construct`
    that creates an async context manager for safely setting up and tearing down
    an instance of that class.
    """

    construct: Callable
```

hexkit/custom_types.py
```diff
@@ -31,13 +31,35 @@
...
 # context manager for safely setting up and tearing down an instance of that class.
 # With the current typing features of Python, it seems not possible to correctly type
 # a class with that signature.
 # E.g. one could define a typing.Protocol subclass with following method:
 # ```
-# class ContextConstructable(Protocol):
+# class AsyncContextConstructable(Protocol):
 #     @classmethod
 #     @asynccontextmanager
 #     def construct(cls, *args: Any, **kwargs: Any): ...
```
Cito, line 65:
```suggestion
#     async def construct(cls, *args: Any, **kwargs: Any): ...
```

hexkit/custom_types.py
```diff
@@ -51,4 +73,4 @@
 #     def construct(cls, foo: str): ...
```
Cito, line 73:
```suggestion
#    async def construct(cls, foo: str): ...
```

hexkit/inject.py
```diff
@@ -29,82 +29,87 @@
...
-    "ContextConstructor",
+    "AsyncConstructor",
     "AsyncInitShutdownError",
     "Configurator",
 ]
 
 
 class NotConstructableError(TypeError):
-    """Thrown when a ContextConstructable expected but not obtained."""
+    """Thrown when a AsyncContextConstructable expected but not obtained."""
```
Cito, line 51:
```suggestion
    """Thrown when an AsyncContextConstructable was expected but not obtained."""
```

hexkit/inject.py
```diff
@@ -29,82 +29,87 @@
...
+def assert_async_constructable(constructable: type[AsyncContextConstructable]):
     """
     Make sure that the provided object has a callable attribute `construct`.
     If this check passes, it can be seen as a strong indication that the provided object
-    is compliant with our definition of a ContextConstructable. However, it does not
-    check whether `construct` really returns an async context manager.
+    is compliant with our definition of a Async(Context)Constructable. However, it does
+    not check whether `construct` really returns an awaitable or an async context
+    manager.
     """
```
Cito, line 66: Maybe use `iscoroutinefunction` here?

tests/fixtures/inject.py
```diff
@@ -65,7 +65,23 @@ def shutdown(self, resource: Resource) -> None:  # type: ignore
         resource.in_context = False
 
 
-class ValidConstructable:
+class ValidAsyncConstructable:
+    """
+    A test class with a `construct` method that is an async context manager.
```
Cito, line 70:
```suggestion
    A test class with an async `construct` method that returns an instance.
```

tests/fixtures/inject.py
```diff
@@ -65,7 +65,23 @@ def shutdown(self, resource: Resource) -> None:  # type: ignore
...
+        return cls(foo=foo)
+
+    def __init__(self, foo: str = "foo"):
+        """Init TestConstructable."""
+        self.foo: Optional[str] = foo
+
+
+class ValidAsyncContextConstructable:
     """
     A test class with a `construct` method that is an async context manager.
```
Cito, line 86:
```suggestion
    A test class with a `construct` method that creates an async context manager.
```

tests/unit/test_inject.py
```diff
@@ -52,74 +54,86 @@
     ],
 )
 async def test_assert_constructable(
-    constructable: ContextConstructable, does_raises: bool
+    constructable: AsyncContextConstructable, does_raises: bool
```
Cito, line 57:
```suggestion
    constructable: Type[AsyncContextConstructable], does_raises: bool
```

tests/unit/test_inject.py
```diff
@@ -52,74 +54,86 @@
...
+            NoAsyncConstructable,
+            None,
+        ),  # passed resources are not initialized in this test
         (object(), NotConstructableError),
         (NoMethodConstructable, NotConstructableError),
     ],
 )
 async def test_context_constructor_init(
-    constructable: ContextConstructable, exception: Optional[type[Exception]]
+    constructable: AsyncContextConstructable, exception: Optional[type[Exception]]
```
Cito, line 82:
```suggestion
    constructable: Type[AsyncContextConstructable], exception: Optional[type[Exception]]
```

tests/unit/test_inject.py
```diff
@@ -52,74 +54,86 @@
...
-        (ValidConstructable, None),
-        (NoCMConstructable, NotConstructableError),
+        (ValidAsyncConstructable, None, False),
+        (ValidAsyncContextConstructable, None, True),
+        (NoAsyncConstructable, NotConstructableError, True),
     ],
 )
 async def test_context_constructor_setup_teardown(
-    constructable: ContextConstructable, exception: Optional[type[Exception]]
+    constructable: AsyncContextConstructable,
```
Cito, line 106:
```suggestion
    constructable: Type[AsyncContextConstructable],
```

Cito (commented):
Useful addition. I have suggested a type for the Constructables that is a little bit better than just Any, can you check that out?

hexkit/custom_types.py
KerstenBreuer: I already tried this some weeks ago but mypy is not recognizing this.
Will push a proof in a minute.

Since we are already making checks for the correct type at runtime, I think there is little value in investing much more time into that right now. If you agree, I can put it on the refactoring list.

hexkit/custom_types.py
KerstenBreuer: I already tried this some weeks ago but mypy is not recognizing this.
Will push a proof in a minute.

Since we are already making checks for the correct type at runtime, I think there is little value in investing much more time into that right now. If you agree, I can put it on the refactoring list.

hexkit/inject.py
KerstenBreuer: That will not match:
````
@asynccontextmanager
async def construct(cls, *args: Any, **kwargs: Any): ...
```

Since construct returns an async context manager and not an awaitable.

hexkit/custom_types.py
KerstenBreuer: Please have a look at this branch: https://github.com/ghga-de/hexkit/pull/36
Where mypy is reporting the following error: https://github.com/ghga-de/hexkit/actions/runs/3248759888/jobs/5330354785#step:6:169

hexkit/custom_types.py
KerstenBreuer: Please have a look at this branch: https://github.com/ghga-de/hexkit/pull/36
Where mypy is reporting the following error: https://github.com/ghga-de/hexkit/actions/runs/3248759888/jobs/5330354785#step:6:169

KerstenBreuer:
Thanks @Cito, could you have another look?

hexkit/custom_types.py
Cito: You can define `ValidAsyncConstructable` as a subclass of `AsyncConstructable` to make it work.

Cito (approved):

KerstenBreuer:
Thanks a lot @Cito

PR #37, opened 2022-10-18
improve serialization of complex model fields

KerstenBreuer:
Complex fields in a DTO model such as datetimes are now more reliably serialized.

Bump version to 0.5.1.


Cito (approved):
Looks good. Checked that the test fails for both the date and the Path without the fix.

PR #38, opened 2022-10-27
raise exceptions when no hits for find_one (GDEV-1406)

KerstenBreuer:
```
The find_one method of DAOs now raises and exception instead of returning None when no hit was found.

Bumps version to 0.6.0.

```

### DIFF
--- a/hexkit/protocols/dao.py
+++ b/hexkit/protocols/dao.py
@@ -83,7 +83,7 @@ class MultipleHitsFoundError(FindError):
 
 
 class NoHitsFoundError(FindError):
-    """Raised when a DAO find operation did result in multiple hits while only a
+    """Raised when a DAO find operation did result in no hits while a
     single hit was expected."""
 
     def __init__(self, *, mapping: Mapping[str, str]):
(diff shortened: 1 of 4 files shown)


hexkit/protocols/dao.py
```diff
@@ -82,6 +82,18 @@ def __init__(self, *, mapping: Mapping[str, str]):
         super().__init__(message)
 
 
+class NoHitsFoundError(FindError):
+    """Raised when a DAO find operation did result in multiple hits while only a
+    single hit was expected."""
```
Cito, line 87:
```suggestion
    """Raised when a DAO find operation did result in no hits while a
    single hit was expected."""
```

hexkit/providers/mongodb/provider.py
```diff
@@ -183,27 +184,24 @@ async def find_one(self, *, mapping: Mapping[str, Any]) -> Optional[Dto]:
...
-                the hist list that was inserted last. Defaults to "single".
 
         Returns:
             Returns a hit in the form of the respective DTO model or None if no hit
             was found.
 
         Raises:
+            NoHitsFoundError:
+                If no hit was found.
             MultpleHitsFoundError:
```
Cito, line 195:
```suggestion
            MultipleHitsFoundError:
```

Cito (approved):
Yes, this is better. One docstring still needed to be adapted.

PR #42, opened 2022-11-08
Feature/multipart copy gdev 1328

mephenor:
```
self._client.copy already implements multipart copy
Adjusted copy_object to use explicit part size

Co-authored-by: Moritz Hahn <Moritz.Hahn@uni-tuebingen.de>
Co-authored-by: KerstenBreuer <kersten-breuer@outlook.com>
```

Manually tested with a 80 GiB and a 250GiB file, seems to work just fine.

### DIFF
--- a/hexkit/protocols/objstorage.py
+++ b/hexkit/protocols/objstorage.py
@@ -19,7 +19,7 @@
 
 import re
 from abc import ABC, abstractmethod
-from typing import NamedTuple, Optional
+from typing import Any, NamedTuple, Optional
 
 __all__ = ["PresignedPostURL", "ObjectStorageProtocol"]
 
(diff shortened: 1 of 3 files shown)


MoritzHahn1337 (approved):

hexkit/protocols/objstorage.py
```diff
@@ -198,6 +198,16 @@ async def get_object_download_url(
             bucket_id=bucket_id, object_id=object_id, expires_after=expires_after
         )
 
+    async def get_object_metadata(
+        self, *, bucket_id: str, object_id: str
+    ) -> dict[str, Any]:
```
KerstenBreuer, line 203: I like this idea, however, the return type should define which data to expect. Or if we currently only want to query the file size, we can also simply name it `get_object_size`. (Or if we currently don't need it as part of the public interface then the _get_object_metadata method is enough, too.) If we keep it as public method, we should also add tests.

KerstenBreuer (changes requested):
Looks good, just the public get_object_metadata method should be adapted.

hexkit/protocols/objstorage.py
mephenor: I'll remove the public-facing part and keep the internals, might need them somewhere else.
Regarding the return type: That's what we get back from the head_object call, so currently I'm just passing it through.
Should we wrap this somehow?

MoritzHahn1337 (approved):

KerstenBreuer (approved):

PR #49, opened 2023-01-31
added method to retrieve object size (GDEV-1567)

KerstenBreuer:
- added object to public interface
- removed _get_object_metadata from protocol

Bumps version to 0.9.0.

### DIFF
--- a/hexkit/providers/s3/provider.py
+++ b/hexkit/providers/s3/provider.py
@@ -777,6 +777,28 @@ class S3ObjectStorage(
 
         return presigned_url
 
+    async def _get_object_size(self, *, bucket_id: str, object_id: str) -> int:
+        """
+        Returns the size of an object in bytes.
+
+        *To be implemented by the provider. Input validation is done outside of this
+        method.*
+        """
+
+        await self._assert_object_exists(bucket_id=bucket_id, object_id=object_id)
+
+        object_metadata = await self._get_object_metadata(
+            bucket_id=bucket_id, object_id=object_id
+        )
+
+        if not "ContentLength" in object_metadata:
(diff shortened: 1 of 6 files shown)


hexkit/providers/s3/provider.py
```diff
@@ -777,6 +777,28 @@ async def _get_object_download_url(
 
         return presigned_url
 
+    async def _get_object_size(self, *, bucket_id: str, object_id: str) -> int:
+        """
+        Returns the size of an object in bytes.
+
+        *To be implemented by the provider. Input validation is done outside of this
+        method.*
```
mephenor, line 785:
```suggestion
```

hexkit/providers/s3/provider.py
```diff
@@ -777,6 +777,28 @@ async def _get_object_download_url(
...
+
+        await self._assert_object_exists(bucket_id=bucket_id, object_id=object_id)
+
+        object_metadata = await self._get_object_metadata(
+            bucket_id=bucket_id, object_id=object_id
+        )
+
+        if not "ContentLength" in object_metadata:
+            raise self.ObjectError(
+                f"Could not get the size of the object with ID '{object_id}' in the"
```
mephenor, line 796:
```suggestion
                f"Could not get the size of the object with ID '{object_id}' in"
```

mephenor (approved):
Could also change https://github.com/ghga-de/hexkit/blob/ee999db656bc1eec0b839e128cdf8edfc69d265d/hexkit/providers/s3/provider.py#L838-L842 to use this now

mephenor (approved):

mephenor (approved):

PR #55, opened 2023-03-14
Update to new way of defining dev requirements

Cito:
- Adapt to new microservices template
- Use requirements file instead of dev extra
- Update pydantic and dev requirements
- Fix some minor issues


setup.cfg
```diff
@@ -56,34 +56,10 @@ s3 =
...
-    types-requests==2.28.11.7
-    mkdocs==1.4.2
-    mkdocs-autorefs==0.4.1
-    mkdocs-material==9.0.3
-    mkdocs-material-extensions==1.1.1
-    mkdocstrings==0.19.1
-    mkdocstrings-python-legacy==0.2.3
-    # 3.4.1 -> 3.4.2 causes issues in hexkit, but 3.7.1 works again
-    # 3.7.1 currently does not work with ghga-connector and dcs
-    testcontainers[kafka,mongo]==3.4.1
```
KerstenBreuer, line 81: Unfortunately, we have to keep at least some of the dependencies in here because they are used by other microservices when using the test hexkit test utils. Maybe we can split this up into: akafka-test, s3-test, and mongodb-test

requirements-dev.txt
```diff
@@ -0,0 +1,6 @@
+# requirements for development and testing this service
+
+-r requirements-dev-common.txt
+
+# additional requirements can be listed here
+testcontainers[kafka,mongo]==3.4.1
```
KerstenBreuer, line 6:
```suggestion
# additional requirements can be listed here
```

KerstenBreuer (changes requested):

KerstenBreuer (approved):
Ja, sieht gut aus.

PR #58, opened 2023-05-25
Added functionality to list all bucket object IDs

mephenor:


hexkit/__init__.py
```diff
@@ -15,4 +15,4 @@
 
 """A Toolkit for Building Microservices using the Hexagonal Architecture"""
 
-__version__ = "0.9.3"
+__version__ = "0.9.4"
```
KerstenBreuer, line 18: This is a new feature, thus maybe bumpt the minor:
```suggestion
__version__ = "0.10.0"
```

KerstenBreuer (changes requested):
Looks good, could you also add a simple test for it?

KerstenBreuer (approved):

PR #60, opened 2023-07-19
Split test fixtures back out (GSI 305)

TheByronHimes:
The previous PR for 0.10.1 imported the S3, MongoDB, and Kafka fixtures into one module for convenience. However, that broke the separation inherent in the package structure. When a service installs hexkit 0.10.1 with less than all three extra dependencies (s3,mongodb,akafka) and tries to use the fixtures, errors will occur because the fixtures.py module in hexkit can't import whatever is missing (or rather, the things it imports in turn rely on packages that won't be there). 
Solution is to get rid of fixtures.py and leave fixture creation in the `testutils` modules.

`mongodb_fixture`, `s3_fixture`, and `kafka_fixture` will be found again in the respective `hexkit.providers.XYZ.testutils` modules. To create a non-function-scoped fixture, you will import `get_mongodb_fixture`, `get_s3_fixture`, or `get_kafka_fixture`, respectively from `hexkit.providers.XYZ.testutils`.

Using a joint fixture with anything other than the default function scope will require overriding the event loop fixture. For convenience, this is available from hexkit:
```
from hexkit.providers.testing.utils import get_event_loop
event_loop = get_event_loop("module")  
```
There are two tests as a sanity check that this is working.


hexkit/providers/akafka/testutils.py
```diff
@@ -408,7 +409,8 @@ async def expect_events(
 
 
 async def kafka_fixture_function() -> AsyncGenerator[KafkaFixture, None]:
-    """Pytest fixture for tests depending on the Kafka-base providers."""
+    """Pytest fixture for tests depending on the Kafka-base providers.
+    **Do not call directly** Instead, use get_kafka_fixture()"""
```
Cito, line 413: Just as a general remark, let's follow [PEP 257](https://peps.python.org/pep-0257/) in docstrings.

"one-line summary, followed by a blank line, followed by a more elaborate description"

The point is that some auto doc creation tools split the docstring in summary and description, which are then displayed differently. The summary should fit into one line.

Cito, line 413:
```suggestion
    """Pytest fixture function for tests depending on the Kafka-base providers.
    
    Do not call directly. Instead, use get_kafka_fixture() to get the actual fixture.
    """
```

Cito (dismissed):
Looks much cleaner to me now, thanks.

hexkit/providers/akafka/testutils.py
TheByronHimes: I had not seen that PEP yet, thanks for bringing it to my attention. I read through it and reformatted the doc strings I touched in this PR.

Cito (approved):

mephenor (approved):

PR #63, opened 2023-08-03
Enable ID in find_all() (GSI-190)

TheByronHimes:
There was a problem with find_all() in the mongodb base DAO where you couldn't include a DTO's id in the mapping because you'd run into one of two problems:
1. You use the id field as it's labeled on the DTO and it doesn't match anything in the database because the ID fields are relabeled "_id" upon insertion.
2. You use "_id" in the mapping to get around the above issue, but this fails because the validation method sees that "_id" is not in the DTO schema. 

We can check the mapping that is passed in to find_all and, if `self._id_field` is present, make a copy of the mapping with the id field renamed to "_id".

### DIFF
--- a/tests/integration/test_mongodb.py
+++ b/tests/integration/test_mongodb.py
@@ -82,6 +82,24 @@ async def test_dao_find_all_with_id(mongodb_fixture: MongoDbFixture):  # noqa: F
     no_results = [x async for x in dao.find_all(mapping={"id": "noresults"})]
     assert len(no_results) == 0
 
+    # make sure other fields beside ID aren't getting ignored
+    no_results_multifield = [
+        x
+        async for x in dao.find_all(
+            mapping={"id": resource_inserted.id, "field_b": 134293487}
+        )
+    ]
+    assert len(no_results_multifield) == 0
+
+    multifield_found = [
+        x
+        async for x in dao.find_all(
+            mapping={"id": resource_inserted.id, "field_b": resource_inserted.field_b}
+        )
+    ]
+    assert len(multifield_found) == 1
+    assert multifield_found[0] == resource_inserted
+
     # find_one calls find_all, so double check that it works there too
     result = await dao.find_one(mapping={"id": resource_inserted.id})
     assert result == resource_inserted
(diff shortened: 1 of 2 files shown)


mephenor (approved):

tests/integration/test_mongodb.py
```diff
@@ -53,6 +53,35 @@ class ExampleDto(ExampleCreationDto):
...
+    resource_inserted = await dao.insert(resource_to_create)
+
+    assert isinstance(resource_inserted, ExampleDto)
+    assert resource_inserted.dict(exclude={"id"}) == resource_to_create.dict()
+
+    # retrieve the resource with find_all
+    resources_read = [
+        x async for x in dao.find_all(mapping={"id": resource_inserted.id})
+    ]
+    assert len(resources_read) == 1
```
Cito, line 78:
```suggestion
    assert len(resources_read) == 1
    assert resources_read[0].id == resource_inserted.id
```

Cito (dismissed):

tests/integration/test_mongodb.py
TheByronHimes: Thanks. I'll also add something to check that you can use id in find_one

tests/integration/test_mongodb.py
```diff
@@ -53,6 +53,40 @@ class ExampleDto(ExampleCreationDto):
...
+    assert resources_read[0].id == resource_inserted.id
+
+    # make sure the previous check wasn't a false positive
+    no_results = [x async for x in dao.find_all(mapping={"id": "noresults"})]
+    assert len(no_results) == 0
+
```
Cito, line 84: Can you also add a test with an additional field like `field_b` to make sure that the other items in the mapping are not ignored when there is an id?

Cito (changes requested):

tests/integration/test_mongodb.py
TheByronHimes: See latest in [1d3bc90](https://github.com/ghga-de/hexkit/pull/63/commits/1d3bc90cea9723d460c96220ed1f1732ea9be06b)

Cito (approved):

mephenor (approved):

PR #66, opened 2023-09-13
Add SSL Support to Kafka Provider Configuration

walzph:
This PR introduces support for SSL communication between our service and Kafka brokers:

- Added new fields `security_protocol` and `ssl_context` to the `KafkaConfig` class. This allows users to specify the desired communication protocol (`PLAINTEXT` or `SSL`) and provide an SSL context if required.
  
- Updated the Kafka producer and consumer initialization code to utilize the new `security_protocol` and `ssl_context` configuration properties. 

These changes ensure that we can establish secure connections to Kafka when necessary, enhancing the security posture of our system.

Bumps version to 0.11.0.


hexkit/providers/akafka/provider.py
```diff
@@ -69,6 +70,16 @@ class KafkaConfig(BaseSettings):
         example=["localhost:9092"],
         description="A list of connection strings to connect to Kafka bootstrap servers.",
     )
+    security_protocol: str = Field(
+        "PLAINTEXT",
```
Cito, line 74: This should be a str literal `Literal["PLAINTEXT", "SSL"]` or a boolean if only these two values are supported. Alternatively, an enum.

hexkit/providers/akafka/provider.py
```diff
@@ -69,6 +70,16 @@ class KafkaConfig(BaseSettings):
...
+        description="Protocol used to communicate with brokers. "
+        + "Valid values are: PLAINTEXT, SSL."
+    )
+    ssl_context: ssl.SSLContext = Field(
```
Cito, line 78: This will be hard to configure. Maybe use the arguments of `create_ssl_context` as configuration options instead.

Cito (changes requested):

hexkit/providers/akafka/provider.py
```diff
@@ -93,6 +119,19 @@ def generate_client_id(*, service_name: str, instance_id: str) -> str:
     """
     return f"{service_name}.{instance_id}"
 
+def generate_ssl_context(config: KafkaConfig) -> Optional[ssl.SSLContext]:
+    """
+    Generate ssl_context for connecting to Kafka broker via an encrypted SSL connection
+    """
+    if config.security_protocol == "SSL":
+        return create_ssl_context(
+            cafile=config.ssl_cafile,  # CA used to sign certificate.
```
Cito, line 128: - You can remove the comments here since the fields are described above.

- Maybe use a ternary expression: `return create_ssl_context(...) if ... else None`

- Do you want to add the tests in a separate PR?

Cito (approved):
Looks good, but see comments.

KerstenBreuer (approved):
Looks nice. Thanks a lot. But as Christoph mentioned, tests would be nice either in this PR or a separate one.

KerstenBreuer:
Ah just saw that mypy and bandit are failing in the static checks.

hexkit/providers/akafka/provider.py
```diff
@@ -149,6 +199,8 @@ async def construct(
 
         producer = kafka_producer_cls(
             bootstrap_servers=",".join(config.kafka_servers),
+            security_protocol=config.security_protocol,
+            ssl_context=generate_ssl_context(config),
```
Cito, line 203: These two args must be also added to the init method of the KafkaProducerCompatible Protocol above. Same for Consumer.

Cito (changes requested):

KerstenBreuer:
Hi @walzph, I will help you migrating to the latest template.

KerstenBreuer (approved):

KerstenBreuer:
Hi @walzph, it is now compatible with the main branch again. Would you like to add tests to this PR or in a separate one?

hexkit/providers/akafka/provider.py
Cito: Ok, I started to rewrite the tests now.

mephenor (approved):

.devcontainer/kafka_secrets/create_secrets.sh
```diff
@@ -0,0 +1,72 @@
...
+if [ -f ca.crt -o -f client.crt ]
+then
+  echo "Secrets already exist. You can remove them to create new ones."
+  exit 0
+fi
+
+# install openssl and keytool
+if ! command -v openssl &> /dev/null || ! command -v keytool &> /dev/null
+then
+  if ! apt install -y openssl openjdk-11-jre-headless
```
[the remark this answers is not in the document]
Cito: No, it does something similar but for keystores which are used in the Java universem and therefore, also in Kakfa.

.devcontainer/kafka_secrets/create_secrets.sh
```diff
@@ -0,0 +1,72 @@
...
+echo $pw > pwd.txt
+
+# generate CA key
+openssl req -new -x509 -keyout ca.key -out ca.crt -days 9999 \
+  -subj '/CN=ca.test.ghga.dev/OU=TEST/O=GHGA' \
+  -passin pass:$pw -passout pass:$pw
+
+for component in kafka client
+do
+  echo "Create keystore for Kafka $component..."
```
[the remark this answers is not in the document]
Cito: Well spotted. Renamed it back to "broker" so that this makes more sense again.

tests/integration/test_akafka.py
```diff
@@ -95,3 +98,75 @@ async def test_kafka_event_subscriber(kafka_fixture: KafkaFixture):  # noqa: F81
...
+
+    This test uses the broker configured with the needed secrets via docker-compose
+    instead of a test container.
+    """
+    payload: JsonObject = {"test_content": "Be aware... Connect with care"}
+    type_ = "test_type"
+    key = "test_key"
+    topic = "test_topic"
+
+    admin_client = KafkaAdminClient(bootstrap_servers=["localhost:9092"])
```
KerstenBreuer, line 131: By convention, we are never using the docker-compose services of the .devcontainer for testing but always use testcontainers. Would you see an option to stick to that convention. Then we also would not have to modify the github actions.

Cito: That's not so easy in this case, let's discuss offline.

.devcontainer/kafka_secrets/create_secrets.sh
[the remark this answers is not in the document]
mephenor: I was trying to figure out, if keytool is actually installed by the command below after the check for its existence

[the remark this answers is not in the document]
Cito: Yes, it's part of  openjdk-11-jre-headless.

KerstenBreuer (approved):
Looks very good, thanks for going the extra mile.

src/hexkit/providers/akafka/containers.py
```diff
@@ -0,0 +1,177 @@
+# Copyright 2021 - 2023 Universität Tübingen, DKFZ, EMBL, and Universität zu Köln
```
KerstenBreuer, line 1: Just maybe rename to testcontainer.py to communicate that this is only for testing.

KerstenBreuer (dismissed):

src/hexkit/providers/akafka/containers.py
Cito: Good idea. Did that now.

Cito: Also updated the verson to 1.0.0. You need to approve again.

KerstenBreuer (approved):

PR #67, opened 2023-09-29
Update from template

mephenor:
This only contains the necessary changes to adapt to the move to ruff.

There were some issues with mypy complaining about stuff that it probably had no problems with before, i.e. not being able to correctly infer information for asyncio.motor stuff and with some BasicConfig in the env vars test.
Hopefully I didn't mute proper issues there.


tests/unit/test_dao.py
```diff
@@ -73,6 +73,11 @@ async def _get_dao(
         """*To be implemented by the provider. Input validation is done outside of this
         method.*
         """
+        ...
+
+        """*To be implemented by the provider. Input validation is done outside of this
```
TheByronHimes, line 78: Is this duplicated on purpose?

src/hexkit/providers/akafka/__init__.py
```diff
@@ -19,7 +19,10 @@
 respectively.
 """
 
-from .provider import KafkaEventPublisher  # noqa: F401
-from .provider import KafkaConfig, KafkaEventSubscriber
+from .provider import (
```
TheByronHimes, line 22: This should be an absolute import. Can be fixed in future PR though, not important.

src/hexkit/providers/mongodb/__init__.py
```diff
@@ -18,6 +18,6 @@
 and associated utilities.
 """
 
-from .provider import MongoDbConfig, MongoDbDaoFactory  # noqa: F401
+from .provider import MongoDbConfig, MongoDbDaoFactory
```
TheByronHimes, line 21: Absolute import here too (unless there's a reason not to do that in an __init__.py)

src/hexkit/providers/mongodb/provider.py
```diff
@@ -419,7 +420,7 @@ def __init__(
         )
         self._db = self._client[self._config.db_name]
 
-    def __repr__(self) -> str:
+    def __repr__(self) -> str:  # noqa: D105
```
TheByronHimes, line 423: Maybe we don't need D105, should consider ignoring it.

src/hexkit/providers/s3/provider.py
```diff
@@ -19,6 +19,8 @@
 Utilities for testing are located in `./testutils.py`.
 """
 
+# ruff: noqa: PLR0913
```
TheByronHimes, line 22: The default max-args setting for ruff.pylint is 5, but the number of ignores here suggest we should maybe increase it.

src/hexkit/providers/s3/provider.py
```diff
@@ -208,35 +204,29 @@ def _translate_s3_client_errors(  # noqa: C901
             exception = cls.ObjectAlreadyExistsError(
                 bucket_id=bucket_id, object_id=object_id
             )
-        elif error_code == "ObjectAlreadyInActiveTierError":
```
TheByronHimes, line 211: Why did this get deleted? Are we not using it anymore?

TheByronHimes (commented):
Just some comments. Looks good but can you tell me about removing the "ObjectAlreadyInActiveTierError" part?

tests/unit/test_dao.py
mephenor: That was already there, fixed that in the other branch and forgot it here...

src/hexkit/providers/s3/provider.py
mephenor: This branch was exactly duplicated directly above

src/hexkit/providers/akafka/__init__.py
mephenor: I did not touch stuff that was already there and was not a direct issue with the move to ruff and pyproject.toml.
If we want to properly vet for relative imports, there are probably at least a few more places and it would make sense to have that change self-contained in its own PR.

src/hexkit/providers/s3/provider.py
TheByronHimes: Okay, it didn't show in the diff (guess I could have viewed the file)

mephenor: Would not have seen that either, if ruff did not complain

tests/unit/test_dao.py
mephenor: Removed in 8a4c959a424537f0e49958f2e10ca1838bbd5701

TheByronHimes (approved):

KerstenBreuer (dismissed):

src/hexkit/providers/mongodb/provider.py
```diff
@@ -59,8 +61,8 @@ def __init__(
         *,
         dto_model: type[Dto],
         id_field: str,
-        collection: AsyncIOMotorCollection,
-        session: Optional[AsyncIOMotorClientSession] = None,
+        collection: AsyncIOMotorCollection,  # type: ignore
+        session: Optional[AsyncIOMotorClientSession] = None,  # type: ignore
```
Cito, line 65: To avoid the error, you can try this:

```python
from motor.core import AgnosticCollection, AgnosticClientSession
```

And then:

```python
        collection: AgnosticCollection,
        session: Optional[AgnosticClientSession] = None,
```

tests/integration/test_akafka_testutils.py
```diff
@@ -35,18 +36,23 @@ def test_delete_topics_specific(kafka_fixture: KafkaFixture):  # noqa: F811
     admin_client = KafkaAdminClient(
         bootstrap_servers=kafka_fixture.config.kafka_servers
     )
+    new_topics = [
+        NewTopic(name="test", num_partitions=1, replication_factor=1),
+        NewTopic(name="test2", num_partitions=1, replication_factor=1),
+    ]
+    admin_client.create_topics(new_topics)
```
Cito, line 43: Next time, better make changes in a separate PR.

Cito (dismissed):
Maybe check that Motor typing issue (see comment below), otherwise looks good to me.

Cito (approved):

Cito (approved):

.github/workflows/pypi_publish.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
 
       - name: Verify Package Version vs Tag Version
         run: |
-          PKG_VER="$(python setup.py -V)"
+          PKG_VER="$(grep -oP 'version = "\K[^"]+' pyproject.toml)"
```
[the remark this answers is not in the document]
Cito: Update: Looks like it actually works with the nested quotes.

Cito (approved):

PR #68, opened 2023-10-10
Upgrade to pydantic v2 (GSI 338)

TheByronHimes:
pydantic v1.10.6 -> 2.4.2
---
I applied the migration tool provided by Pydantic [here](https://docs.pydantic.dev/2.4/migration/#code-transformation-tool).

### BaseSettings
The biggest change for us is that `BaseSettings` is no longer part of pydantic, but rather pydantic_settings. 
There are also some changes to how the configuration is done for the BaseSettings classes. For example, the Config override is gone and replace with a property named `model_config`, which should be a `SettingsConfigDict`. 
The `customise_sources` override is now called `settings_customise_sources`. 

### BaseModel
The most impactful changes in `hexkit` with regard to BaseModels are:
- The `.dict()` method is deprecated and replaced with `.model_dump()`. By default this will output regular python objects, but if passed `mode='json'`, it will only output JSON-serializable values in the dict.
- Validators are done differently. In `hexkit` we were only using them in the FileObject for testing, and really only to set values. Instead, you can/must use `field_validators`. In this specific case I used the `computed_field` decorator because the values for `content` and `md5` were really being _computed_ from the value of `file_path`.
- The same config changes are here too, so any instance of `class Config():` should be replaced with `model_config = ConfigDict(...)`. Note that for BaseModel it's ConfigDict and for BaseSettings it's SettingsConfigDict. 
- The `.schema()` method which returns the fields of the model is now replaced with `.model_json_schema()`.
- `.json()` is deprecated, should now use `.model_dump_json()`.
- `.copy()` is now `.model_copy()`.


KerstenBreuer (approved):
Very nice thanks a lot

src/hexkit/providers/s3/testutils.py
```diff
@@ -61,21 +61,21 @@ class FileObject(BaseModel):
...
-    content: bytes = b"will be overwritten"
-    md5: str = "will be overwritten"
-
-    # pylint: disable=no-self-argument
-    @validator("content", always=True)
-    def read_content(cls, _, values):  # noqa: N805
-        """Read in the file content."""
-        with open(values["file_path"], "rb") as file:
+
+    @computed_field  # type: ignore [misc]
```
Cito, line 65: Good idea to use a computed field here.

src/hexkit/providers/s3/testutils.py
```diff
@@ -61,21 +61,21 @@ class FileObject(BaseModel):
...
+    @property
+    def content(self) -> bytes:
+        """Extract the content from the file at the provided path"""
+        if not self.file_path:
+            return b"will be overwritten"
```
Cito, line 70: Wouldn't it better better to raise an error in this case?

Cito (approved):
Thanks a lot for working on this, looks good to me. See some small remarks.

src/hexkit/providers/s3/testutils.py
TheByronHimes: That's not a bad idea. changed in the commit below.

Cito (approved):

PR #69, opened 2023-10-30
Add s3 provider test files as package data 

dontseyit:
This PR updates pyproject.toml to include test files that should be distributed with the package. They were probably excluded by mistake during the template update, this configuration adds them back. 

Since these are the only YAML files required are specified as recommended here https://setuptools.pypa.io/en/latest/userguide/datafiles.html#subdirectory-for-data-files


KerstenBreuer (approved):
I am approving this so that you are unblocked and can release these changes.

However, could you please have a look at this old PR:
https://github.com/ghga-de/hexkit/pull/65/files#diff-41d5a52589e0480be9c099d2bba7a8135b8b0d71bcbb8df3582a8df1c2295003R1

Using MANIFEST.in file is build system independent.

Would you mind switching to that in a separate PR?

PR #73, opened 2023-11-22
Fix lower version limit for testcontainers

Cito:


Cito:
The Kafka SSL container does not work with older versions of testcontainers.

TheByronHimes (approved):

PR #74, opened 2023-11-30
Correlation id utilities (GSI 501)

TheByronHimes:
This adds the correlation ID utilities as outlined in the Marabou Stork epic.
This PR only adds the utilities and does not modify the behavior of the Kafka providers (that will come next). 
The new stuff:

### Functions
- `get_correlation_id()`
- `new_correlation_id()`
- `set_correlation_id()`
- `validate_correlation_id()`
- `set_context_var()`

### Errors
- `CorrelationIdContextError`
- `InvalidCorrelationIdError`
- `MissingCorrelationIdError`
I think `MissingCorrelationIdError` could be folded into the `InvalidCorrelationIdError` as it's only used in `set_correlation_id` to differentiate between an empty string vs an invalid non-empty string.

### Other
- `correlation_id_var`: The ContextVar instance that is used to store/access a correlation ID for a context.


mephenor (dismissed):
Looks nice. Personally I'd fold the CorrelationIDContext and MissingContextID errors into one, don't see a reason to keep them separate if we don't necessarily need to distinguish between unset and empty ID.

mephenor (approved):

tests/unit/test_correlation.py
```diff
@@ -0,0 +1,141 @@
...
+    with pytest.raises(exception) if exception else nullcontext():
+        validate_correlation_id(correlation_id)
+
+
+@pytest.mark.parametrize(
+    "correlation_id,exception",
+    [
+        ("12345", InvalidCorrelationIdError),
+        ("", InvalidCorrelationIdError),
+        (new_correlation_id(), None),
```
Cito, line 89: Creating the new id is actually also non-deterministic, but I think we can live with it here.

Cito: But maybe also add a fixed valid value before that line.

tests/unit/test_correlation.py
```diff
@@ -0,0 +1,141 @@
...
+        async with set_correlation_id(correlation_id=correlation_id):
+            pass
+
+
+@pytest.mark.parametrize(
+    "correlation_id,exception",
+    [
+        ("12345", InvalidCorrelationIdError),
+        ("", CorrelationIdContextError),
+        (new_correlation_id(), None),
```
Cito, line 107: same here

tests/unit/test_correlation.py
```diff
@@ -0,0 +1,141 @@
...
+    test_var: ContextVar[str] = ContextVar("test", default=default)
+
+    # Make sure the initial `get()` returns the default value
+    assert test_var.get() == default
+
+    # Ensure the value is set in the context manager
+    async with set_context_var(test_var, outer_value):
+        assert test_var.get() == outer_value
+
+        # Ensure the value that is reset is actually the previous token, not just default
```
Cito, line 135:
```suggestion
        # Ensure the value that is reset is actually the previous value, not just default
```

Cito (changes requested):
Looks good, just some suggestions regarding the testing.

TheByronHimes:
@Cito please see the changes in [24dce0f](https://github.com/ghga-de/hexkit/pull/74/commits/24dce0fe7e31864ac45782008bb99f6cc2cdd236) and resolve the conversations if you are satisfied. I'm happy to make further modifications if needed.

Cito (approved):
Thanks for the changes.

PR #76, opened 2023-12-07
Correlation ID in Kafka Providers (GSI 506)

TheByronHimes:
This PR introduces a backwards-compatible way to use correlation IDs with Kafka in hexkit.
The configuration addition is the same as the `ghga-service-commons` change -- a `generate_correlation_id` flag that will set the correlation ID in the event header if it doesn't exist in the environment. 

### Publishing ⬆️ 
The `_publish_validated()` method has been modified to retrieve the cid from the ContextVar. If it's empty and the generate flag is set, it will create a new cid to use for the header. Otherwise it will use the validated cid (or raise an error if it's not valid). 

### Consuming ⬇️ 
The `_consume_event()` method has been modified to retrieve the cid from the event header. Since the approach is the same as the way the 'type' was previously extracted, the logic was abstacted into a separate method. If the header doesn't exist, the event is simply ignored. If the header is present but invalid, an error will be raised.

### Other 🦦 
Pytest-asyncio v0.23.0 introduced a new way to work with non-function-scoped async fixtures. The event loop override is now out, and a simple keyword for @pytest.mark.asyncio is in. To avoid ballooning the scope of this PR, I've capped pytest-asyncio at < 0.23.0 for now (0.22.x was yanked). I think a fix for this will require a major version bump in hexkit.

### DIFF
--- a/src/hexkit/providers/akafka/provider.py
+++ b/src/hexkit/providers/akafka/provider.py
@@ -82,7 +82,7 @@ class KafkaConfig(BaseSettings):
     kafka_ssl_cafile: str = Field(
         default="",
         description="Certificate Authority file path containing certificates"
-        + " used to sign broker certificates. If a CA not specified, the default"
+        + " used to sign broker certificates. If a CA is not specified, the default"
         + " system CA will be used if found by OpenSSL.",
     )
     kafka_ssl_certfile: str = Field(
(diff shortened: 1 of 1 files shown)


Cito (approved):
Looks good to me, very nice. Also like your detailed PR descriptions.

I also noticed the event loop changes in `pytest_asyncio` in one of the services and also solved this by capping the version. Should be properly solved in a separate PR.

mephenor (approved):
Looks good

PR #77, opened 2023-12-12
Tools for Structured Logging (GSI 510)

TheByronHimes:
This PR introduces a way to achieve consistently structured, JSON-formatted logs. 

>**_TL;DR_**
>1. Add LoggingConfig to Config() as subclass, 
>2. Define logger objects as usual in modules
>3. In top-level module: `from hexkit.log_tools import configure_logging` 
>4. Use `configure_logging` with loggers as needed, but if using module name for logger name, should suffice to only configure the parent logger.
>5. Output is in JSON, and you can override the JSON format with your own format string by configuring log_format.

### Config
There is a new config class, `LoggingConfig`, which contains:
- `log_level`
- `service_name`
- `service_instance_id`
- `log_format`

The service name and instance ID will therefore be shared with the KafkaConfig class. The log_level is the same parameter as in the `ApiConfigBase` class, except that the values are upper case because the native python logging library expects capitalized values. (The next step would be to remove the log_level value from ApiConfigBase completely since it won't be needed in the app config step anymore.)

**Using the Config:**
The LoggingConfig class should be added to the usual `Config` class in services. In the top-level module, such as `main.py`, the config instance should be passed to `configure_logging()` for each logger to configure. 

### Log Output
The default output format is a one-line JSON string (from json.dumps()). When using the JSON format, the logs will always contain the following structure:
```
{
  "timestamp": ...,
  "service": ...,
  "instance": ...,
  "level": ...,
  "correlation_id": ...,
  "message": ...,
  "details": {
    ...
  }
}
```

To add items to `details`, use the "extra" keyword in log calls. Example:
```python
log.info("The file with ID %s is already in the inbox.", file_id, extra={"file_id": file_id})
```
would result in a "details" section like this:
```
{
  ...,
  "details": {
    "file_id": "123xyz"
  }
}
```

To override the JSON formatting, you must supply a formatting string for `log_format`. 
When using log_format, the JsonFormatter object is replaced with the basic Formatter, 
and the log_format string is passed as-is. There is currently no validation performed on log_format,
so it is possible to provide a broken format string and get log errors as a result.

As an implementation detail, there is a RecordCompiler class that subclasses StreamHandler. Its purpose is to put together all the non-standard attributes listed above before passing them to the formatter.


tests/unit/test_logging.py
```diff
@@ -0,0 +1,273 @@
...
+
+DEFAULT_CONFIG = LoggingConfig(service_name="", service_instance_id="")
+
+
+@pytest.fixture(autouse=True)
+def reset_config():
+    """Reset the config before and after each test."""
+    LoggerFactory.configure(log_config=DEFAULT_CONFIG)
+    yield
+    LoggerFactory.configure(log_config=DEFAULT_CONFIG)
```
mephenor, line 69: If this is used by all tests, it should probably suffice to just configure once in the beginning

mephenor (changes requested):
Really like the thorough explanation, made it easier to get some of the missing context, even though I still had to look up the details for StremHandler and LoggerAdapter to make proper sense of things. There's two minor issues, but apart from that it looks solid.

Regarding reconfiguration: The major use cases would be initial configuration in a service and custom logging in tests, as far as I can see, and reconfiguring in the actual code should be avoided to have consistent output, correct?

KerstenBreuer (commented):
Great solution. I added some suggestions in #79 that might simplify it a bit. Happy to discuss

KerstenBreuer:
Hi Byron, could you maybe merge the v2 branch into this one so that it only shows what has actually changed?

TheByronHimes:
> Hi Byron, could you maybe merge the v2 branch into this one so that it only shows what has actually changed?

Done

src/hexkit/log.py
```diff
@@ -0,0 +1,160 @@
...
+        record.service = self._service_name
+        record.instance = self._service_instance_id
+        record.level = log_record["levelname"]
+        record.correlation_id = correlation_id_var.get(None)
+        record.details = extras
+        return super().handle(record)
+
+
+def configure_logging(*, logger: Logger, config: LoggingConfig):
+    """Set up logging"""
```
KerstenBreuer, line 152: Actually, it might be nicer to use the root logger by default, because this is most likely what we want to do to enable logging done by other libraries:
```suggestion
def configure_logging(*, config: LoggingConfig, logger: Optional[Logger] = None):
    """Configure logging for the specified logger. If no logger is provided, the root logger is use."""
    if not logger:
        logger = logging.getLogger()
    
```

src/hexkit/log.py
```diff
@@ -0,0 +1,160 @@
...
+    "RecordCompiler",
+    "configure_logging",
+]
+
+
+DUMMY_RECORD = LogRecord("dummy", 0, "dummy", 0, None, None, None, None, None)
+RESERVED_RECORD_KEYS = DUMMY_RECORD.__dict__.keys()
+
+# Add TRACE log level
+addLevelName(5, "TRACE")
```
KerstenBreuer, line 50: Do we really need this additional level (e.g. for uvicorn) or would the default level be sufficient?

tests/unit/test_logging.py
```diff
@@ -0,0 +1,224 @@
...
+    "level",
+    "correlation_id",
+    "details",
+)
+
+DEFAULT_CONFIG = LoggingConfig(service_name="", service_instance_id="")
+
+
+@dataclass
+class JsonLog:
```
KerstenBreuer, line 64: You could also use a TypedDict then you wouldn't need the additional method.

tests/unit/test_logging.py
```diff
@@ -0,0 +1,224 @@
...
+    config = LoggingConfig(
+        log_level="CRITICAL",  # set to higher than default of WARNING
+        service_name=VALID_SERVICE_NAME,
+        service_instance_id=VALID_INSTANCE_ID,
+        log_format="",
+    )
+
+    configure_logging(logger=logger, config=config)
+
+    assert not caplog.records
```
KerstenBreuer, line 116: I thought the caplog fixture was not working propperly?

tests/unit/test_logging.py
```diff
@@ -0,0 +1,224 @@
...
+
+VALID_SERVICE_NAME = "test"
+VALID_INSTANCE_ID = "001"
+VALID_CONFIG = LoggingConfig(
+    log_level="INFO",
+    service_name="hexkit-test",
+    service_instance_id="0",
+)
+
+DEFAULT_KEYS = (  # all the keys present in the resulting log
```
KerstenBreuer, line 40: You could use the RESERVED_RECORD_KEYS from the log module instead.

tests/unit/test_logging.py
```diff
@@ -0,0 +1,224 @@
...
+    assert json_log.instance == config.service_instance_id
+    assert json_log.message == "should log"
+
+
+def test_record_compiler(caplog):
+    """Unit test for the RecordCompiler class."""
+    record_compiler = RecordCompiler(config=DEFAULT_CONFIG)
+
+    assert not caplog.records
+    log = logging.getLogger()  # getting root logger, not configured
```
KerstenBreuer, line 137: Won't using the root logger have side effects on other tests?

tests/unit/test_logging.py
```diff
@@ -0,0 +1,224 @@
...
+
+    assert len(caplog.records) == 1
+    record = caplog.records[0]
+
+    for key in ADDED_KEYS:
+        assert key not in record.__dict__
+
+    # Feed the record to the RecordCompiler
+    record_compiler.handle(record)
+    for key in DEFAULT_KEYS:
```
KerstenBreuer, line 149:
```suggestion
    for key in ADDED_KEYS:
```

src/hexkit/log.py
```diff
@@ -0,0 +1,160 @@
...
+    service_instance_id: str = Field(
+        ...,
+        examples=["germany-bw-instance-001"],
+        description=(
+            "A string that uniquely identifies this instance across all instances of"
+            + " this service. This is included in log messages."
+        ),
+    )
+    log_format: str = Field(
+        default="",
```
KerstenBreuer, line 74:
```suggestion
    log_format: Optional[str] = Field(
        None,
```

tests/unit/test_logging.py
```diff
@@ -0,0 +1,224 @@
...
+        assert key in json_log
+
+    # make sure the message is what we expected
+    assert json_log["message"] == "This is a test"
+
+
+@pytest.mark.parametrize(
+    "log_format, formatter_class",
+    [
+        ("", JsonFormatter),
```
KerstenBreuer, line 188:
```suggestion
        (None, JsonFormatter),
```

tests/unit/test_logging.py
```diff
@@ -0,0 +1,224 @@
...
+        ("%(message)s", logging.Formatter),
+    ],
+)
+def test_formatter_selection(log_format: str, formatter_class: type[logging.Formatter]):
```
KerstenBreuer, line 192:
```suggestion
def test_formatter_selection(log_format: Optional[str], formatter_class: type[logging.Formatter]):
```

KerstenBreuer (approved):
Just some minor suggestions.

tests/unit/test_logging.py
TheByronHimes: It intercepts logs before formatting. It's fine for checking if a log has been emitted

tests/unit/test_logging.py
TheByronHimes: The DEFAULT_KEYS lists the final fields included in the JSON log, which are different from the built-in attributes found in RESERVED_RECORD_KEYS. Maybe it should be renamed for clarity

tests/unit/test_logging.py
TheByronHimes: In general it could, but it shouldn't in this case. I'll rename them anyway.

tests/unit/test_logging.py
KerstenBreuer: Ah OK, I see

tests/unit/test_logging.py
KerstenBreuer: Ah OK, sorry missunderstood

src/hexkit/log.py
Cito: TRACE is even more fine grained and verbose than DEBUG and normally used to get output from system calls or 3rd party libs. As far as I see it's supported by uvicorn to log on the level of ASGI events. It does not harm if we support it as well, though normally INFO or DEBUG should be sufficient to catch bugs.

src/hexkit/log.py
Cito: Better keep the `default=` since otherwise mypy and pyright have problems.

src/hexkit/log.py
TheByronHimes: I agree that it would be easier to get 3rd party loggers to work that way. Please see [a4e2c2d](https://github.com/ghga-de/hexkit/pull/77/commits/a4e2c2da6dca7ccd3144a9aa436aa266cadc1d00) for relevant changes

src/hexkit/log.py
TheByronHimes: Thanks, I've made the change in [95ad845](https://github.com/ghga-de/hexkit/pull/77/commits/95ad845d31aed82970ecc469a48c74ea70978a29)

tests/unit/test_logging.py
TheByronHimes: See above

tests/unit/test_logging.py
TheByronHimes: Changes in [95ad845](https://github.com/ghga-de/hexkit/pull/77/commits/95ad845d31aed82970ecc469a48c74ea70978a29)

src/hexkit/log.py
KerstenBreuer: OK, didn't knew that this is causing issues.

tests/unit/test_logging.py
```diff
@@ -0,0 +1,237 @@
...
+    )
+
+    configure_logging(config=config, logger=log)
+    assert log.getEffectiveLevel() == trace
+    assert isinstance(log.handlers[0].formatter, logging.Formatter)
+
+
+def test_root_config():
+    """Test that the root logger is configured by default."""
+    root = logging.getLogger()
```
KerstenBreuer, line 230: Resetting the root logger after the test might be good to avoid side effects. I.e. removing the added handler.

KerstenBreuer (changes requested):
Just found this one last thing

src/hexkit/log.py
Cito: The main issue is that you get errors when you're instantiation the model without specifying that field. We often silence these with comments but it's not necessary if you define the field with the default keyword.

tests/unit/test_logging.py
TheByronHimes: Yeah probably a good call, [ba5e705](https://github.com/ghga-de/hexkit/pull/77/commits/ba5e705c4d75189e51c386659846c717933b2643) should take care of that.

tests/unit/test_logging.py
```diff
@@ -0,0 +1,244 @@
...
+    for handler in root.handlers:
+        assert not isinstance(handler, RecordCompiler)
+
+    configure_logging(config=DEFAULT_CONFIG)
+
+    assert any([isinstance(handler, RecordCompiler) for handler in root.handlers])
+
+    for handler in root.handlers:
+        if isinstance(handler, RecordCompiler):
+            root.removeHandler(handler)
```
KerstenBreuer, line 242: Might still be a problem if the test fails before cleanup. This should fix or use a fixture:
```suggestion
    try:
        configure_logging(config=DEFAULT_CONFIG)
        root_handlers = list(root.handlers).copy()
    finally:
        for handler in root.handlers:
            if isinstance(handler, RecordCompiler):
                root.removeHandler(handler)

    assert any([isinstance(handler, RecordCompiler) for handler in root_handlers])
```

TheByronHimes: I went with a fixture and put it in the `utils` module so it can be used by other tests if needed. [5194b2f](https://github.com/ghga-de/hexkit/pull/77/commits/5194b2f77a0239dd88ff4432d4f6161f7cd89a71)

KerstenBreuer (approved):

PR #79, opened 2023-12-14
Suggestions

KerstenBreuer:
- avoids the factory pattern
- allows to configure only once at the top of the package
- no configuration needed for each module
- also works for code not written by us, such as third party libraries
- added example

Kept the original suggestion as is. Tests missing.

### DIFF
--- a/examples/log_example/log_example/some_module.py
+++ b/examples/log_example/log_example/some_module.py
@@ -1,0 +1,29 @@
+# Copyright 2021 - 2023 Universität Tübingen, DKFZ, EMBL, and Universität zu Köln
+# for the German Human Genome-Phenome Archive (GHGA)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""Some module"""
+
+import logging
+
+from hexkit.correlation import new_correlation_id, set_correlation_id
+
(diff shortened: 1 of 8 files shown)


src/hexkit/log.py
```diff
@@ -0,0 +1,159 @@
...
+        record.instance = self._service_instance_id
+        record.level = log_record["levelname"]
+        record.correlation_id = correlation_id_var.get(None)
+        record.details = extras
+        return super().handle(record)
+
+
+def configure_logging(*, logger: Logger, config: LoggingConfig):
+    """Set up logging"""
+    formatter = JsonFormatter()
```
TheByronHimes, line 152: `formatter = Formatter(config.log_format) if config.log_format else JsonFormatter()`
Or we can do 
`formatter = JsonFormatter(config.log_format)` and handle the logic in the JsonFormatter class (I think `super().format(record)`)

examples/log_example/log_example/some_module.py
```diff
@@ -0,0 +1,29 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""Some module"""
+
+import logging
+
+from hexkit.correlation import new_correlation_id, set_correlation_id
+
```
TheByronHimes, line 22: `from hexkit.log_tools import configure_logging`

examples/log_example/log_example/some_module.py
```diff
@@ -0,0 +1,29 @@
...
+logger = logging.getLogger(__name__)
+
```
TheByronHimes, line 24: configure_logging(logger, Config())

examples/log_example/log_example/some_module.py
```diff
@@ -0,0 +1,29 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""Some module"""
+
+import logging
+
```
TheByronHimes, line 20: `from my_service.config import Config`

TheByronHimes:
Thanks Kersten, I appreciate the constructive feedback. This is a clean improvement.

PR #80, opened 2023-12-18
Upgrade pytest-asyncio (GSI-534)

TheByronHimes:
Bumps pytest-asyncio to version 0.23.2 and removes the event_loop fixture. 
Instead of overriding the event loop, you now pass the 'scope' kwarg to @pytest.mark.asyncio when using non-function-scoped fixtures.


Cito (approved):
Looks good to me.

PR #85, opened 2023-12-22
fix to the latest template

KerstenBreuer:
regarding dev_install, list_outdated_dependencies, and pyproject generation.


TheByronHimes:
@KerstenBreuer The dockerfile in the root dir needs to be updated as well

TheByronHimes (approved):

PR #88, opened 2024-01-22
Update pypi publish

TheByronHimes:
The line `PKG_VER="$(grep -oP 'version = "\K[^"]+' pyproject.toml)"` was became broken when we added the tooling config to pyproject.toml because the regex then matched on `target-version = "py39"` (for ruff) as well as `minversion = "7.1"` (for pytest).

To fix this, I've added `^` before `version` so it will only match the desired line.

### DIFF
--- a/.github/workflows/pypi_publish.yaml
+++ b/.github/workflows/pypi_publish.yaml
@@ -18,17 +18,8 @@ jobs:
 
       - name: Ensure package version and tag are equal
         run: |
-          if [ -f "setup.py" ]
-          then
-            PKG_VER="$(python setup.py --version)"
-          elif [ -f "pyproject.toml" ]
-          then
-            version=$(python3 -c "import tomli; print(tomli.load(open('pyproject.toml', 'rb'))['project']['version'])")
-            PKG_VER=$version
-          else
-            echo "Neither setup.py nor pyproject.toml were found." >&2; exit 1
-          fi
 
+          PKG_VER="$(grep -oP '^version = "\K[^"]+' pyproject.toml)"
           TAG_VER="${GITHUB_REF##*/}"
 
           echo "Package version is $PKG_VER" >&2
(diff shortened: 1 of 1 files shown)


.github/workflows/pypi_publish.yaml
```diff
@@ -56,7 +70,7 @@ jobs:
           pip install
           -r requirements-dev.txt
 
-      - name: Run pytest on freshly installed package
+      - name: Run pytest on freshly install package
```
mephenor, line 73: Just nitpicking, but this should stay `installed`, right?

.github/workflows/pypi_publish.yaml
```diff
@@ -16,10 +16,24 @@ jobs:
...
+        shell: bash
         run: |
-          PKG_VER="$(grep -oP 'version = "\K[^"]+' pyproject.toml)"
-          TAG_VER="${GITHUB_REF##*/}"
+          if [ -f "setup.py" ]
+          then
+            PKG_VER="$(python setup.py --version)"
+          elif [ -f "pyproject.toml" ]
+          then
+            version=$(python3.11 -c "import tomllib; print(tomllib.load(open('pyproject.toml', 'rb'))['project']['version'])")
```
mephenor, line 28:
```suggestion
            version=$(python3 -c "import tomli; print(tomli.load(open('pyproject.toml', 'rb'))['project']['version'])")
```
We set up python with version 3.9 above, no need to hardcode a version number here.
Which also means, we don't have tomllib available, as that's only part of the stdlib from 3.11 onward.
Alternatively, you  could install tomli, as that's what we use in e.g. the `update_pyproject` script

mephenor (changes requested):
The pyproject part should be handled differently if we don't want to explicitly rely on python 3.11.

Not sure about the inputs, someone else might need  to chime in on those.

mephenor (approved):
Looks good now.
Could you also integrate the change into the template respository?

PR #89, opened 2024-01-23
Reference requirements in lock folder 

TheByronHimes:
Reference to the requirements-dev.txt file was out of date, now that the file has been moved to /lock.


mephenor (approved):
Here we go again...

PR #90, opened 2024-01-24
Consistent spelling of correlation ID in logs

Cito:
When filtering out log messages about correlation IDs in an integration test, I was bitten by the fact that in one case a different spellings of "id" was used (lower instead upper case). The PR changes just this log message so make it easier to filter correlation ID related messages.


TheByronHimes (approved):

TheByronHimes (commented):
Do you need to bump the version and do a release?

PR #91, opened 2024-02-01
Switch event subscriber consumers to manual commit mode

mephenor:
The default consumer now commits after `consume_event` is finished and no longer periodically in the background.
Added a test case to verify this new behavior and bumped version number.

A `ConsumerRebalanceListener` has not been implemented, as `on_partitions_revoked(self, revoked)` only provides a list of  `TopicPartition` that are currently assigned to the consumer. 
For the rebalance listener to work, a consumer has to be used with the `subscribe` method and passed a concrete instance of a `ConsumerRebalanceListener`, so all this is an implementation detail of a consumer.
The easiest solution to implement that without breaking our current abstractions would be subclassing the AioKafkaConsumer.

This would also need a bit more investigation, as there are some details that are not fully clear.
In the simplest case, the rebalance listener would just wait for all ongoing consumption to finish, by comparing local offsets. This should be possible, as no new events are fetched once `on_partitions_revoked(self, revoked)`  has been called. However, this might block the rebalancing up to a configured timeout and we might need to fine tune the rebalance timeout parameter, if we ever have long running jobs or accumulate a backlog. 

For now, we probably could get away with wrapping `event_subscriber.run` in a try-except block to gracefully handle commit failures.


.pyproject_generation/pyproject_custom.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "hexkit"
-version = "2.1.0"
+version = "2.2.0"
```
KerstenBreuer, line 3: I think it is rather a bugfix, right? We though it would work like that in the first place. Now we match the expectations.

src/hexkit/providers/akafka/provider.py
```diff
@@ -473,3 +479,4 @@ async def run(self, forever: bool = True) -> None:
         else:
             event = await self._consumer.__anext__()
             await self._consume_event(event)
+        await self._consumer.commit()
```
KerstenBreuer, line 482: This would just commit once the consumer is stopped, right? But we would like to commit after every successfully consumed event. Would modifiying the _consumer_event method be suitable:

```
    async def _consume_event(self, event: ConsumerEvent) -> None:
        """Consume an event by passing it down to the translator via the protocol."""
        event_label = self._get_event_label(event)
        headers = headers_as_dict(event)
        try:
            type_ = get_header_value(header_name="type", headers=headers)
            correlation_id = get_header_value(
                header_name="correlation_id", headers=headers
            )
        except EventHeaderNotFoundError as err:
            logging.warning("Ignored an event: %s. %s", event_label, err.args[0])
            return
        if type_ in self._types_whitelist:
            logging.info('Consuming event of type "%s": %s', type_, event_label)
            try:
                async with set_correlation_id(correlation_id):
                    # blocks until event processing is completed:
                    await self._translator.consume(
                        payload=event.value,
                        type_=type_,
                        topic=event.topic,
                    )
            except Exception:
                logging.error(
                    "A fatal error occurred while processing the event: %s",
                    event_label,
                )
                raise
            else:
                await self._consumer.commit()
        else:
            logging.info("Ignored event of type %s: %s", type_, event_label)```

File: tests/integration/test_akafka.py [hunk 3]
```diff
@@ -162,3 +166,56 @@ async def test_kafka_ssl(tmp_path: Path):
...
+        translator=translator,
+    ) as event_subscriber:
+        # consume one event:
+        await event_subscriber.run(forever=False)
+
+        # provide correct type information
+        consumer = cast(AIOKafkaConsumer, event_subscriber._consumer)
+
+        # check that the consumer is in manual commit mode
+        assert not consumer._enable_auto_commit
```
KerstenBreuer, line 206: We should test behavior and not implementation, can probably be skipped:
```suggestion
```

KerstenBreuer (changes requested):
I agree, the rebalancing is not needed at the moment.
Would be great to add a test that confirms that an event is re-consumed if a failure occurred while consuming it the first time.

src/hexkit/providers/akafka/provider.py
mephenor: Yep, put that in the wrong place.
All the places that ignore the event also need a commit, else we constantly reconsume events that are not even processed.

mephenor:
Regarding the test: I'll move some of the test code in the IRS exploration branch to hexkit, that should do the trick.

src/hexkit/providers/akafka/provider.py
KerstenBreuer: Yes right

KerstenBreuer (dismissed):

mephenor:
Made another small change:
If the consumer is reused after an exception, it potentially hangs indefinitely on the `__anext__()` call without reconsuming.
As an exception in the EventSubscriber will the crash the respective service, at least for now, this shouldn't be a problem in practice. 

If we decide to reuse an EventSubscriber in such an event, it (or at least the associated consumer) probably needs to be restarted to work properly, but I haven't tested if this actually works that way.

KerstenBreuer (approved):

PR #92, opened 2024-02-05
db event outbox (GSI-610)

KerstenBreuer:
```
Implements a wrapper around the existing DaoFactoryProtocol that adds outbox behavior for publishing.
An additional __metadata__ field is added to documents in MongoDB that encodes whether a change has
been published and whether a document was deleted or not.
It is assumed that one topic only contains events related to one resource type stored in one collection.

Bumps version to 2.2.0.
```


TheByronHimes (approved):
Very nice, looks like an elegant solution.

Cito (dismissed):
That should work.

We can also consider using dates instead of booleans.

mephenor (approved):
Not much to add to the previous reviewers.
I like the idea

Cito (approved):

KerstenBreuer:
Implements a wrapper around the existing DaoFactoryProtocol that adds outbox behavior for publishing.
An additional __metadata__ field is added to documents in MongoDB that encodes whether a change has
been published and whether a document was deleted or not.
It is assumed that one topic only contains events related to one resource type stored in one collection.

Bumps version to 2.2.0.

PR #102, opened 2024-04-11
Speed up tests and support Python 3.12

Cito:
This PR contains the following changes:

- speed up test suite (about 5x) by reusing test containers 🚀
- remove the deprecated `delete_topics` method
- remove the `kafka-python` dependency (which is not supported on Python 3.12)
- tests with Python 3.12 using tox

Since we now have to test with multiple Python versions, a faster test suite is even more important.

The following things should still be done before releasing this as 3.0.0:

- ~add Python 3.12 as supported version in pyproject.toml~
  (done via service template)
- the tox tests should also run on GitHub with all Python versions
  (this is a bit difficult, see issue GSI-712)


src/hexkit/providers/akafka/testutils.py
```diff
@@ -408,71 +406,50 @@ def _build_record_deletion_command(self, delete_config: JsonObject) -> str:
...
             if topics is None:
-                topics = [
-                    topic for topic in all_topics if topic != "__consumer_offsets"
-                ]
+                if exclude_internal:
+                    topics = [
+                        topic
+                        for topic in await admin_client.list_topics()
+                        if not topic.startswith("__")
+                    ]
```
mephenor, line 428:
```suggestion
                    ]
                else:
                    topics = await admin_client.list_topics()
```
Docstring is ambiguous here for  `exclude_internal == False`.
Currently topics would be `None` in this case.
Is this intended?

mephenor (changes requested):
Nice readability improvements.

There's one refactored piece of code that needs clarification if what it's doing is intentional.

src/hexkit/providers/akafka/testutils.py
Cito: Yes, the admin client understands that None = all.

Cito: Added a comment to explain this now.

mephenor (dismissed):

TheByronHimes (approved):

PR #106, opened 2024-05-06
Fix pyenv issues and some more (GSI-750)

Cito:
- Export classes from mongokafka
- Update lock files and hook revs
- Use Python 3.12 as the system Python
- Fix issues caused by ruff (the scripts need to be formatted for Python 3.12 because they are provided by the template, the rest for Python 3.9 for backward compatibility of the lib)
- Bump version to 3.0.1


scripts/check_license.py
```diff
@@ -34,45 +33,45 @@
...
-    ".pylintrc",
-    ".ruff.toml",
-    ".ruff_cache",
-    "example_config.yaml",
-    "config_schema.json",
-    "LICENSE",  # is checked but not for the license header
-    ".pre-commit-config.yaml",
-    "docs",
-    ".vscode",
+    ".gitignore" ".pre-commit-config.yaml",
```
TheByronHimes, line 43:
```suggestion
    ".gitignore", ".pre-commit-config.yaml",
```

TheByronHimes (approved):

scripts/check_license.py
Cito: This should be fixed in the service template.

PR #111, opened 2024-06-04
Allow selective suppression of publication in DaoPublisher

Cito:
DaoPublisher: Allow `dto_to_event` to return None and in that case suppress the publication of the event.


TheByronHimes (approved):
Very nice

PR #112, opened 2024-06-05
Iron out flaky kafka test (GSI-793)

TheByronHimes:
The `.getmany()` call seems to often pull one record at a time in this test, at least when run locally. During the time the tests were failing, there was a `break` statement that would get hit after retrieving any records, regardless of count.

### DIFF
--- a/tests/integration/test_akafka_testutils.py
+++ b/tests/integration/test_akafka_testutils.py
@@ -136,9 +136,8 @@ async def test_clear_topics_specific(kafka: KafkaFixture):
 
     # make sure messages are consumed again
     records = []
-    while (fetches := 0) < 10:
+    for _ in range(10):
         # use limit in case of changes that cause unintended infinite loop here
-        fetches += 1
         prefetched = await consumer.getmany(timeout_ms=500)
         if prefetched:
             records.extend(next(iter(prefetched.values())))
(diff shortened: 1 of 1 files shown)


tests/integration/test_akafka_testutils.py
```diff
@@ -136,15 +136,17 @@ async def test_clear_topics_specific(kafka: KafkaFixture):
 
     # make sure messages are consumed again
     records = []
-    while True:
-        prefetched = await consumer.getmany(timeout_ms=5000)
-        if not prefetched:
+    while (fetches := 0) < 10:
```
Cito, line 139: `for _ in range(10):` is more idiomatic.

tests/integration/test_akafka_testutils.py
```diff
@@ -136,15 +136,17 @@ async def test_clear_topics_specific(kafka: KafkaFixture):
...
+        # use limit in case of changes that cause unintended infinite loop here
+        fetches += 1
+        prefetched = await consumer.getmany(timeout_ms=500)
+        if prefetched:
+            records.extend(next(iter(prefetched.values())))
+        if len(records) >= 2:
             break
```
Cito, line 146: We could put something like `wait asyncio.sleep(0.05)` here at the very end of the loop.

Cito (dismissed):
Looks like a good solution for this problem, thanks,

tests/integration/test_akafka_testutils.py
TheByronHimes: I changed it in [b61524f](https://github.com/ghga-de/hexkit/pull/112/commits/b61524f6ee2630c742a9e0bca67352005e6a0a60)

Cito (approved):
The final commit was the proper fix, thanks for taking the time to solve this!

PR #113, opened 2024-06-11
Minor improvements regarding correlation IDs

Cito:
- Add default message to CorrelationIdContextError (for cases where only the error message itself is logged)
- Export util function to set a new correlation ID (for running code outside a request)


src/hexkit/correlation.py
```diff
@@ -80,6 +84,16 @@ async def set_correlation_id(correlation_id: str):
...
 
 
+@asynccontextmanager
+async def set_new_correlation_id():
+    """Set a new correlation ID for the life of the context."""
+    correlation_id = new_correlation_id()
+
+    async with set_context_var(correlation_id_var, correlation_id):
+        log.info("Set context correlation ID to %s", correlation_id)
+        yield
```
TheByronHimes, line 94: If you changed this to `yield correlation_id`, would that allow you to update tests to reference the new value directly instead of making a subsequent call to `get_correlation_id()`? So in tests it would then be:
```python
async with set_new_correlation_id() as temp_correlation_id:
    assert temp_correlation_id != correlation_id
```

TheByronHimes (dismissed):
Thanks for adding that!

src/hexkit/correlation.py
Cito: Good idea, implemented in 0a28adb now.

TheByronHimes (approved):

PR #115, opened 2024-06-27
Refactor S3 Testutils (GSI-762)

TheByronHimes:
Bumps version from 3.2.2 to 3.2.3 because the changed items were not in use by any other microservices and are low-impact. Strictly following SemVer would mean bumping the major version number to 4 because the `delete_created_buckets` method was deleted, and some other functions were moved into the s3 fixture instead of existing on their own (and another was deleted).

The fixtures were moved into `s3/testutils/_fixures.py`, the supporting functions moved into `s3/testutils/_utils.py`, and `typical_workflow` was moved into its own module in that same directory. The imports are all funneled into `__init__.py` to avoid disrupting consuming services.
A list of current buckets is retrieved using the object storage's s3 client. The bucket deletion process iterates through the buckets and deletes them one by one, so neither the object storage nor test fixture need to track which buckets have been created.


src/hexkit/providers/s3/testutils/_fixtures.py
```diff
@@ -0,0 +1,302 @@
...
+    "persistent_s3_fixture",
+    "get_clean_s3_fixture",
+    "clean_s3_fixture",
+    "s3_fixture",
+    "tmp_file",
+    "temp_file_object",
+]
+
+
+LOCALSTACK_IMAGE = "localstack/localstack:3.4.0"
```
Cito, line 67:
```suggestion
LOCALSTACK_IMAGE = "localstack/localstack:3.5.0"
```

We can use this PR to update the image.

src/hexkit/providers/s3/testutils/_fixtures.py
```diff
@@ -0,0 +1,302 @@
...
+
+    def __init__(self, config: S3Config, storage: S3ObjectStorage):
+        """Initialize with config."""
+        self.config = config
+        self.storage = storage
+
+    def get_buckets(self) -> set[str]:
+        """Return a list of the buckets currently existing in the S3 object storage."""
+        response = self.storage._client.list_buckets()
+        buckets = set(bucket["Name"] for bucket in response["Buckets"])
```
Cito, line 99:
```suggestion
        buckets = {bucket["Name"] for bucket in response["Buckets"]}
```

src/hexkit/providers/s3/testutils/_fixtures.py
```diff
@@ -0,0 +1,302 @@
...
+            # Delete all of these objects
+            for object_id in object_ids:
+                await self.storage.delete_object(bucket_id=bucket, object_id=object_id)
+
+    async def delete_buckets(self, buckets_to_exclude: Optional[list[str]] = None):
+        """Delete the populated buckets."""
+        buckets = self.get_buckets()
+        for bucket in list(buckets.difference(buckets_to_exclude or [])):
+            await self.storage.delete_bucket(bucket, delete_content=True)
+            buckets.discard(bucket)
```
Cito, line 128:
```suggestion
        for bucket in self.get_buckets().difference(buckets_to_exclude or []):
            await self.storage.delete_bucket(bucket, delete_content=True)
```

Cito (approved):
Thanks a lot, this is pretty much what I had in mind.

Just small suggestions.

src/hexkit/providers/s3/testutils/_fixtures.py
TheByronHimes: Good point. Fixed this and the below comments

PR #116, opened 2024-07-01
Expose event header for testing (GSI-770)

TheByronHimes:
Right now it's difficult to verify the correlation ID when verifying the rest of a published event in testing, because the context disappears as soon as the consume action is complete. You have to set up a dummy translator or other workaround.

This PR adds a `capture_headers: bool = False` parameter to `EventRecorder`, `KafkaFixture.expect_events`, and `KafkaFixture.record_events`.  
The default allows for backwards compatibility. When left False, the `RecordedEvent`s returned or examined by the above tools will effectively ignore the event headers.  
If the flag is set to True, then the `RecordedEvent`s will contain the event headers in a new field called...headers.  
The ExpectedEvent class also got the optional `headers` field, of course.

But the automatic comparison that occurs in, e.g. `KafkaFixture.expect_events`, will only compare the headers between an expected and a recorded event if the expected event was formed with headers specified. This also preserves backwards compatibility. 

A test has been added to examine the behavior for the matrix of combinations for the following:
- `capture_headers` (True/False)
- Expected headers set (Not set/Value 1/Value 2)
- Recorded headers (Value 1/Value 2) (the "None" case is effectively covered when capture_headers is False)

A pre-existing test for the event recorder was updated to test with capture_headers set/not set as well.

Minor version number bumped (3.2.3 --> 3.3.0)

There was some weird reformatting happening from ruff with f-strings. Took the easy road to solve that for now, will have to figure out what's actually happening in the meantime.


tests/integration/test_akafka_testutils.py
```diff
@@ -407,3 +450,75 @@ async def test_expect_events_mismatch(
...
+    # 1. Expected headers are not provided
+    # 2. Capture headers flag is set and expected/recorded headers match
+    error = True
+    if not expected_headers or (
+        capture_headers
+        and (expected_headers.get("correlation_id", "") == recorded_correlation_id)
+    ):
+        error = False
+
+    with pytest.raises(ValidationError) if error else nullcontext():
```
Cito, line 523:
```suggestion
    # only time there should be an error is when 1. expected headers are provided
    # and 2. capture headers flag aren't set or expected/recorded headers don't match
    expected_error = expected_headers and (
        not capture_headers
        or expected_headers.get("correlation_id", "") != recorded_correlation_id
    )

    with pytest.raises(ValidationError) if expected_error else nullcontext():
```

Cito (approved):
Looks good to me. Just some minor code style suggestions.

tests/integration/test_akafka_testutils.py
TheByronHimes: Thanks, this is a little easier to read. I added this and the previous changes.

Cito (approved):

PR #117, opened 2024-07-05
Include traceback in logs (GSI-884)

Cito:
Our custom JSON Log handler that is used as the default log formatter in hexkit currently ignores potentially existing exception info that is normally output as a traceback.

This PR adds the exception info using an additional property `exception` in the JSON output which is an object with the three properties `type`, `message` and `traceback`.

While usually "flat is better than nested", we already use the nested `details` object, so it seems consequential to make the exception info a nested object as well.

Note that this exception info needs to be explicitly passed using the `exc_info` argument of the log function, and normally we don't do this. However, it is done by uvicorn which catches all unhandled exceptions and sets `exc_info` when it logs them.


ckaipf (approved):

lkuchenb (approved):
LGTM

mephenor (approved):

PR #118, opened 2024-07-08
Improve topic clearing (GSI-885)

Cito:
The method to clear topics in the Kafka Fixture used the Kafka CLI tools so far.

This was problematic in the docker compose environment, where you cannot simply run the Kafka CLI in the Kafka container as you can do with the test container.

It was probably done that way because so far there was no better way to delete messags available. However, a week ago aiokafka was released in version 0.11.0 which implements the DeleteRecords API, so that it is now possible to delete message using the admin client.

This PR changes the implementation of the `clear_topics` method to use this new feature in favor of running the Kafka CLI tools.


dontseyit:
The update look good to me. Do we have unit tests to cover this?

dontseyit (approved):

PR #119, opened 2024-07-16
Improve methods to cleanup collections and buckets (GSI-902)

Cito:
PR #115 changed the behavior of the `empty_buckets` method in an attempt to make it more similar to `empty_collections`.

Before, it used to only empty the buckets that were populated using the fixture. Now, it empties all existing buckets that can be accessed. This has the advantage that the fixture does not need to track which buckets were populated and can also clean buckets that were not populated using the fixture. However, this can now also empty buckets in the S3 instance that are not part of the application or tests and not supposed to be touched.

Therefore, this PR adds parameters for specifying the collections and buckets to be emptied in addition to the parameter that already allowed specifying which collections and buckets should be excluded.

In addition, these parameters are now made keyword-only to avoid mistakes when calling these methods.

Another question in this context was how to handle objects that were not created using hexkit and therefore can have names that do not comply with the validation rules for ids implemented in hexkit. These will cause validation errors when hexkit tries to delete them when emptying the containing bucket. This could happen for instance with object names that contain forward slashes or start with a dot. In these cases, we intentionally let the error bubble up instead of trying to delete these objects despite the error or silently ignore them,  because they usually indicate a problem that should be made apparent.

However, this PR also facilitates using different validation rules for object names, by not hard coding the regular expression used for validation, but exposing it as an attribute that can be overridden. Also, the PR relaxes the validation a little bit by allowing underscores and longer object names. Forward slashes are still not allowed by default, though.


src/hexkit/protocols/objstorage.py
```diff
@@ -466,43 +469,43 @@ def _validate_bucket_id(cls, bucket_id: str):
...
     def _validate_object_id(cls, object_id: str):
-        """Check whether a object id follows the recommended naming pattern.
+        """Check whether an object id follows the recommended naming pattern.
         This is roughly based on (plus some additional restrictions):
         https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html
         Raises ObjectIdValidationError if not valid.
+
+        Note that the default implementation does not allow object IDs to contain
+        forward slashes (/) which are sometimes used to represent directories.
+        Providers can overwrite this method or the above constants to allow slashes,
```
TheByronHimes, line 497:
```suggestion
        Providers can override this method or the above constants to allow slashes,
```

src/hexkit/providers/s3/testutils/_fixtures.py
```diff
@@ -111,19 +110,66 @@ async def populate_file_objects(self, file_objects: list[FileObject]):
...
         )
 
-    async def empty_buckets(self, buckets_to_exclude: Optional[list[str]] = None):
-        """Clean the test artifacts or files from the populated buckets."""
-        for bucket in self.get_buckets().difference(buckets_to_exclude or []):
+    async def empty_buckets(
+        self,
+        *,
+        buckets: Optional[Union[str, list[str]]] = None,
+        exclude_buckets: Optional[list[str]] = None,
```
TheByronHimes, line 117:
```suggestion
        exclude_buckets: Optional[Union[str, list[str]]] = None,
```
Either this type should be updated to include `str`, or the `isinstance(exclude_buckets, str)` check should be deleted below

src/hexkit/providers/s3/testutils/_fixtures.py
```diff
@@ -111,19 +110,66 @@ async def populate_file_objects(self, file_objects: list[FileObject]):
...
 
-    async def delete_buckets(self, buckets_to_exclude: Optional[list[str]] = None):
-        """Delete the populated buckets."""
-        for bucket in self.get_buckets().difference(buckets_to_exclude or []):
-            await self.storage.delete_bucket(bucket, delete_content=True)
+    async def delete_buckets(
+        self,
+        *,
+        buckets: Optional[Union[str, list[str]]] = None,
+        exclude_buckets: Optional[list[str]] = None,
```
TheByronHimes, line 152: Same as above

src/hexkit/providers/s3/testutils/_fixtures.py
```diff
@@ -111,19 +110,66 @@ async def populate_file_objects(self, file_objects: list[FileObject]):
...
+    ):
+        """Delete the given bucket(s).
+
+        If no buckets are specified, all existing buckets will be deleted,
+        therefore we recommend to always specify a bucket list.
```
TheByronHimes, line 157: If there's really a threat that buckets outside of test could be unintentionally deleted, maybe it would be better to mandate a bucket list every time.

tests/integration/test_mongodb.py
```diff
@@ -126,17 +126,22 @@ async def test_dao_find_all_without_collection(mongodb: MongoDbFixture):
...
+    mongodb.empty_collections(exclude_collections=["test2"])
+    assert db.list_collection_names() == ["test2"]
+    mongodb.empty_collections()
+    assert db.list_collection_names() == []
 
 
 async def test_dao_happy(mongodb: MongoDbFixture):
     """Test the happy path of performing basic CRUD database interactions using
-    the MongoDbDaoFactory in a surrograte ID setting.
+    the MongoDbDaoFactory in a surrogate ID setting.
```
TheByronHimes, line 144: grate catch

TheByronHimes (changes requested):
Looks good on the whole. I only saw a few small things.

src/hexkit/providers/s3/testutils/_fixtures.py
Cito: Improve type hints

src/hexkit/providers/s3/testutils/_fixtures.py
Cito: In our usual environment (integration tests using test containers) this is not a problem. So I would not want to make it too restrictive.

PR #123, opened 2024-07-29
Kafka DLQ 

TheByronHimes:
>  _UPDATE August 1, 2024_: Services implementing the `EventPublisherProtocol` will have to add the `headers` parameter to `_publish_validated`. Currently, this only affects the [auth-service](https://github.com/ghga-de/auth-service/blob/cd4dbb13146c6356a242b9abf6fbb4c2c09a217f/tests/unit/user_management/user_registry/test_event_pub.py#L55), [access-request-service](https://github.com/ghga-de/access-request-service/blob/26cb9eca1441a4185c06a5a4ccb339043b1a36c0/tests/test_event_pub.py#L43), and [metldata](https://github.com/ghga-de/metldata/blob/daed73aa5ad601e2c0589fc2c68ce69a0e07ae3d/src/metldata/event_handling/event_handling.py#L101).

This PR adds a _not-quite_ backwards-compatible way to handle (retry, dismiss) Kafka events that fail at some point while being processed by the translator. Some config options are added to manage this functionality, and in their default state they change nothing: events that fail will crash the service and be retried upon restart. 

Changes introduced are summarized below:

**New Config Fields:**
- `kafka_dlq_topic`: Name of the topic for events when they initially fail
- `kafka_retry_topic`: Name of the topic for failed events that are to be requeued
- `kafka_max_retries`: The number of times to retry a failed event before sending it to the DLQ topic, if enabled
- `kafka_enable_dlq`: Whether or not to use the DLQ
- `kafka_retry_backoff`: The number of seconds to wait between "immediate" retries. This value is doubled with each subsequent retry.

**New Functionality:**
- Retry events after a pause. Configuring `kafka_max_retries` will set the number of times to directly retry a failed event after waiting `kafka_retry_backoff` seconds. The default number of retries is 0, which will propagate the error just like `hexkit <4`. If `kafka_max_retries` is greater than 0, final failure will result in a `RetriesExhaustedError`.
- Publish failed events to a dead letter queue automatically. Set `kafka_dlq_topic`, `kafka_retry_topic`, and `kafka_enable_dlq` to use the DLQ functionality and publish failed events to the configured DLQ topic. The original topic name is preserved by appending it to the **event** ~~payload~~ **headers** in a field named `_original_topic`. When an event is consumed from the configured `kafka_retry_topic`, its headers will be inspected to find the original topic (so the service knows how to treat the event). For events in `kafka_retry_topic`, having no `_original_topic` header will get them ignored. The same goes for `kafka_dlq_topic` if using the default `process_dlq_event` function.
- Ignore (discard) events in a DLQ topic or send them to a retry topic. The `KafkaDLQSubscriber` class is a dedicated provider that listens to the configured DLQ topic. When run, the event is sent to the configured retry topic or, if `ignore=True` is used, the event is simply ignored. This class does not have to be used by a particular service, and could be used by a dedicated DLQ triage service. Additionally, there is a `process_dlq_event` parameter that takes a callable. This callback is executed when `ignore=False`, and can be a simple standalone function or a method on a complex, configurable class instance. Through `process_dlq_event`, events can be modified or examined as needed before being republished or discarded.
- Encapsulate the way we use `ConsumerEvent` details with `ExtractedEventInfo` (open to naming suggestions). The class will contain the type_, topic, key, payload, and dict-form decoded headers from `ConsumerEvent`, and can be instantiated both with kwargs and a `ConsumerEvent`-compatible object.

> [!NOTE]
> The immediate retry and dlq mechanisms are independent. Events can be published to the dead letter queue with 0 retries, retried a few times and never sent to the DLQ, etc.

## Adoption
For existing services, no changes are needed beyond adjusting the configuration parameters and adding a `KafkaEventPublisher` to the `KafkaEventSubscriber.construct()` or `KafkaOutboxSubscriber.construct()` calls (e.g. in the injection module). For deployment, at least one DLQ and one Retry topic will need to be established (maybe one each for all services is enough to begin with?)

To add the ability to move events from a DLQ topic to a Retry topic, there needs to be a way to invoke a `KafkaDLQSubscriber`, as illustrated below. We need to discuss how best to do this: a dedicated DLQ service? CLI command on each kafka-consuming service? Where applicable, events can be manually sent to a retry queue with Kafka UI as long as `_original_topic` is placed in the headers.

## Example Usage for DLQ Subscriber and Process Function
```python
from typing import Optional

from hexkit.providers.akafka.provider import (
    ConsumerEvent,
    ExtractedEventInfo,
    KafkaDLQSubscriber,
    KafkaEventPublisher,
    validate_dlq_headers,
)

from ifrs.config import Config

# This is an example of a callback function that can be passed to the KafkaDLQSubscriber
# Here it's very simple, but it could also be a complex function on a config-driven class.
async def fix_dlq_event(event: ConsumerEvent) -> Optional[ExtractedEventInfo]:
    """Update old events for config changes or pass event unchanged."""
    validate_dlq_headers(event)
    event_info = ExtractedEventInfo(event)
    if event.topic == "uploads" and event_info.headers["type"] == "outdated_type":
        event_info.headers["type"] = "new_type"
    return event_info

# This is an example of an entrypoint function that could be triggered for DLQ event resolution
async def resolve_dlq_event(ignore: bool = False):
    """Requeue or discard a DLQ event."""
    config = Config()
    async with (
        KafkaEventPublisher.construct(config=config) as publisher,
        KafkaDLQSubscriber.construct(
            config=config, dlq_publisher=publisher, process_dlq_event=fix_dlq_event
        ) as dlq_subscriber,
    ):
        await dlq_subscriber.run(ignore)
```

### DIFF
--- a/src/hexkit/providers/akafka/config.py
+++ b/src/hexkit/providers/akafka/config.py
@@ -117,6 +117,15 @@ class KafkaConfig(BaseSettings):
         title="Kafka Enable DLQ",
         examples=[True, False],
     )
+    kafka_retry_backoff: NonNegativeInt = Field(
+        default=0,
+        description=(
+            "The number of seconds to wait before retrying a failed event. The backoff"
+            + " time is doubled for each retry attempt."
+        ),
+        title="Kafka Retry Backoff",
+        examples=[0, 1, 2, 3, 5],
+    )
 
     @model_validator(mode="after")
     def validate_retry_topic(self):
--- a/src/hexkit/providers/akafka/config.py
+++ b/src/hexkit/providers/akafka/config.py
@@ -86,12 +86,14 @@ class KafkaConfig(BaseSettings):
         default="",
         description="The name of the service-specific topic used for the dead letter queue.",
         examples=["dcs-dlq", "ifrs-dlq", "mass-dlq"],
+        title="Kafka DLQ Topic",
     )
     kafka_retry_topic: str = Field(
         default="",
         description=(
             "The name of the service-specific topic used to retry previously failed events."
         ),
+        title="Kafka Retry Topic",
         examples=["dcs-dlq-retry", "ifrs-dlq-retry", "mass-dlq-retry"],
     )
     kafka_max_retries: NonNegativeInt = Field(
@@ -100,6 +102,7 @@ class KafkaConfig(BaseSettings):
             "The maximum number of times to immediately retry consuming an event upon"
             + " failure. Works independently of the dead letter queue."
         ),
+        title="Kafka Max Retries",
         examples=[0, 1, 2, 3, 5],
     )
     kafka_enable_dlq: bool = Field(
@@ -111,6 +114,7 @@ class KafkaConfig(BaseSettings):
             + " exhausting all retries, and both `kafka_dlq_topic` and"
             + " `kafka_retry_topic` must be set."
         ),
+        title="Kafka Enable DLQ",
         examples=[True, False],
     )
 
--- a/src/hexkit/providers/akafka/config.py
+++ b/src/hexkit/providers/akafka/config.py
@@ -16,7 +16,7 @@
 
 """Apache Kafka specific configuration."""
 
-from typing import Literal, Self
+from typing import Literal
 
 from pydantic import Field, NonNegativeInt, PositiveInt, SecretStr, model_validator
 from pydantic_settings import BaseSettings
@@ -115,7 +115,7 @@ class KafkaConfig(BaseSettings):
     )
 
     @model_validator(mode="after")
-    def validate_retry_topic(self) -> Self:
+    def validate_retry_topic(self):
         """Ensure that the retry topic is not the same as the DLQ topic."""
         if self.kafka_retry_topic and self.kafka_retry_topic == self.kafka_dlq_topic:
             raise ValueError(
--- a/src/hexkit/providers/akafka/config.py
+++ b/src/hexkit/providers/akafka/config.py
@@ -16,9 +16,9 @@
 
 """Apache Kafka specific configuration."""
 
-from typing import Literal
+from typing import Literal, Self
 
-from pydantic import Field, PositiveInt, SecretStr
+from pydantic import Field, NonNegativeInt, PositiveInt, SecretStr, model_validator
 from pydantic_settings import BaseSettings
 
 
@@ -26,12 +26,12 @@ class KafkaConfig(BaseSettings):
     """Config parameters needed for connecting to Apache Kafka."""
 
     service_name: str = Field(
-        ...,
+        default=...,
         examples=["my-cool-special-service"],
         description="The name of the (micro-)service from which messages are published.",
     )
     service_instance_id: str = Field(
-        ...,
+        default=...,
         examples=["germany-bw-instance-001"],
         description=(
             "A string that uniquely identifies this instance across all instances of"
@@ -40,7 +40,7 @@ class KafkaConfig(BaseSettings):
         ),
     )
     kafka_servers: list[str] = Field(
-        ...,
+        default=...,
         examples=[["localhost:9092"]],
         description="A list of connection strings to connect to Kafka bootstrap servers.",
     )
@@ -82,3 +82,49 @@ class KafkaConfig(BaseSettings):
         + " services that have a need to send/receive larger messages should set this.",
         examples=[1024 * 1024, 16 * 1024 * 1024],
     )
+    kafka_dlq_topic: str = Field(
+        default="",
+        description="The name of the service-specific topic used for the dead letter queue.",
+        examples=["dcs-dlq", "ifrs-dlq", "mass-dlq"],
+    )
+    kafka_retry_topic: str = Field(
+        default="",
+        description=(
+            "The name of the service-specific topic used to retry previously failed events."
+        ),
+        examples=["dcs-dlq-retry", "ifrs-dlq-retry", "mass-dlq-retry"],
+    )
+    kafka_max_retries: NonNegativeInt = Field(
+        default=0,
+        description=(
+            "The maximum number of times to immediately retry consuming an event upon"
+            + " failure. Works independently of the dead letter queue."
+        ),
+        examples=[0, 1, 2, 3, 5],
+    )
+    kafka_enable_dlq: bool = Field(
+        default=False,
+        description=(
+            "A flag to toggle the dead letter queue. If set to False, the service will"
+            + " crash upon exhausting retries instead of publishing events to the DLQ."
+            + " If set to True, the service will publish events to the DLQ topic after"
+            + " exhausting all retries, and both `kafka_dlq_topic` and"
+            + " `kafka_retry_topic` must be set."
+        ),
+        examples=[True, False],
+    )
+
+    @model_validator(mode="after")
+    def validate_retry_topic(self) -> Self:
+        """Ensure that the retry topic is not the same as the DLQ topic."""
+        if self.kafka_retry_topic and self.kafka_retry_topic == self.kafka_dlq_topic:
+            raise ValueError(
+                "kafka_retry_topic and kafka_dlq_topic cannot be the same."
+            )
+        if self.kafka_enable_dlq and not (
+            self.kafka_dlq_topic and self.kafka_retry_topic
+        ):
+            raise ValueError(
+                "Both kafka_dlq_topic and kafka_retry_topic must be set when the DLQ is enabled."
+            )
+        return self
--- a/src/hexkit/providers/akafka/provider/daosub.py
+++ b/src/hexkit/providers/akafka/provider/daosub.py
@@ -112,7 +112,7 @@ class KafkaOutboxSubscriber(InboundProviderBase):
         *,
         config: KafkaConfig,
         translators: Sequence[DaoSubscriberProtocol],
-        publisher: Optional[EventPublisherProtocol] = None,
+        dlq_publisher: Optional[EventPublisherProtocol] = None,
         kafka_consumer_cls: type[KafkaConsumerCompatible] = AIOKafkaConsumer,
     ):
         """Setup and teardown an instance of the provider.
@@ -121,8 +121,8 @@ class KafkaOutboxSubscriber(InboundProviderBase):
         - `config`: MongoDB-specific config parameters.
         - `translators`: A sequence of translators implementing the
             `DaoSubscriberProtocol`.
-        - `publisher`: An instance of the publisher to use for the DLQ. Can be None if
-            not using the dead letter queue.
+        - `dlq_publisher`: An instance of the publisher to use for the DLQ. Can be None
+            if not using the dead letter queue. It is used to publish events to the DLQ.
         - `kafka_consumer_cls`: The Kafka consumer class to use. Defaults to
             `AIOKafkaConsumer`.
 
@@ -131,7 +131,7 @@ class KafkaOutboxSubscriber(InboundProviderBase):
         """
         translator_converter = TranslatorConverter(translators=translators)
 
-        if config.kafka_enable_dlq and publisher is None:
+        if config.kafka_enable_dlq and dlq_publisher is None:
             error = ValueError("A publisher is required when the DLQ is enabled.")
             logging.error(error)
             raise error
@@ -139,7 +139,7 @@ class KafkaOutboxSubscriber(InboundProviderBase):
         async with KafkaEventSubscriber.construct(
             config=config,
             translator=translator_converter,
-            publisher=publisher,
+            dlq_publisher=dlq_publisher,
             kafka_consumer_cls=kafka_consumer_cls,
         ) as event_subscriber:
             yield cls(event_subscriber=event_subscriber)
--- a/src/hexkit/providers/akafka/provider/daosub.py
+++ b/src/hexkit/providers/akafka/provider/daosub.py
@@ -21,6 +21,7 @@
 import logging
 from collections.abc import Sequence
 from contextlib import asynccontextmanager
+from typing import Optional
 
 from aiokafka import AIOKafkaConsumer
 from pydantic import ValidationError
@@ -31,6 +32,7 @@ from hexkit.protocols.daosub import (
     DaoSubscriberProtocol,
     DtoValidationError,
 )
+from hexkit.protocols.eventpub import EventPublisherProtocol
 from hexkit.protocols.eventsub import EventSubscriberProtocol
 from hexkit.providers.akafka.config import KafkaConfig
 from hexkit.providers.akafka.provider.eventsub import (
@@ -85,8 +87,8 @@ class TranslatorConverter(EventSubscriberProtocol):
                 dto = translator.dto_model.model_validate(payload)
             except ValidationError as error:
                 message = (
-                    f"The event of type {type_} on topic {topic} was not valid wrt. the"
-                    + " DTO model."
+                    f"The event of type {type_} on topic {topic}"
+                    + " was not valid wrt. the DTO model."
                 )
                 logging.error(message)
                 raise DtoValidationError(message) from error
@@ -110,21 +112,34 @@ class KafkaOutboxSubscriber(InboundProviderBase):
         *,
         config: KafkaConfig,
         translators: Sequence[DaoSubscriberProtocol],
+        publisher: Optional[EventPublisherProtocol] = None,
         kafka_consumer_cls: type[KafkaConsumerCompatible] = AIOKafkaConsumer,
     ):
         """Setup and teardown an instance of the provider.
 
         Args:
-            config: MongoDB-specific config parameters.
+        - `config`: MongoDB-specific config parameters.
+        - `translators`: A sequence of translators implementing the
+            `DaoSubscriberProtocol`.
+        - `publisher`: An instance of the publisher to use for the DLQ. Can be None if
+            not using the dead letter queue.
+        - `kafka_consumer_cls`: The Kafka consumer class to use. Defaults to
+            `AIOKafkaConsumer`.
 
         Returns:
             An instance of the provider.
         """
         translator_converter = TranslatorConverter(translators=translators)
 
+        if config.kafka_enable_dlq and publisher is None:
+            error = ValueError("A publisher is required when the DLQ is enabled.")
+            logging.error(error)
+            raise error
+
         async with KafkaEventSubscriber.construct(
             config=config,
             translator=translator_converter,
+            publisher=publisher,
             kafka_consumer_cls=kafka_consumer_cls,
         ) as event_subscriber:
             yield cls(event_subscriber=event_subscriber)
--- a/src/hexkit/providers/akafka/provider/eventsub.py
+++ b/src/hexkit/providers/akafka/provider/eventsub.py
@@ -389,9 +389,9 @@ class KafkaEventSubscriber(InboundProviderBase):
         correlation_id = event.headers.get("correlation_id", "")
         errors = []
         if not event.type_:
-            errors.append("type_ is empty")
+            errors.append("event type is empty")
         elif event.type_ not in self._types_whitelist:
-            errors.append(f"type_ '{event.type_}' is not in the whitelist")
+            errors.append(f"event type '{event.type_}' is not in the whitelist")
         if not correlation_id:
             errors.append("correlation_id is empty")
         if event.topic in (self._retry_topic, self._dlq_topic):
--- a/src/hexkit/providers/akafka/provider/eventsub.py
+++ b/src/hexkit/providers/akafka/provider/eventsub.py
@@ -28,13 +28,7 @@ import ssl
 from collections.abc import Awaitable
 from contextlib import asynccontextmanager
 from dataclasses import dataclass
-from typing import (
-    Callable,
-    Literal,
-    Optional,
-    Protocol,
-    TypeVar,
-)
+from typing import Callable, Literal, Optional, Protocol, TypeVar, cast
 
 from aiokafka import AIOKafkaConsumer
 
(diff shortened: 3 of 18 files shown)


tests/fixtures/utils.py
```diff
@@ -49,3 +50,58 @@ def root_logger_reset():
...
+) -> str:
+    """Assert that a log message was logged with the given level and message.
+
+    Args:
+    - `level`: The log level to check for.
+    - `message`: The message to check for.
+    - `records`: The log records to check (usually `caplog.records`)
+    - `parse`: Whether to parse the message with the arguments from the log record.
+
+    If a match is found, the parsed message is return (regardless of the value of `parse`).
```
Cito, line 72:
```suggestion
    If a match is found, the parsed message is returned (regardless of the value of `parse`).
```

tests/unit/test_dlqsub.py
```diff
@@ -0,0 +1,598 @@
...
+            dlq_topic=dlq_topic,
+            max_retries=max_retries,
+            enable_dlq=enable_dlq,
+        )
+
+
+@pytest.mark.asyncio()
+async def test_original_topic_is_preserved(kafka: KafkaFixture):
+    """Ensure the original topic is preserved when it reaches the DLQ subscriber and
+    when it comes back to the Retry subscriber.
```
Cito, line 233:
```suggestion
    when it comes back to the subscriber.
```

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -177,76 +210,210 @@ async def construct(
...
+            )
+        except Exception:
+            logging.warning(
+                "Failed initial attempt to consume event of type '%s' on topic '%s' with key '%s'.",
+                event.type_,
+                event.topic,
+                event.key,
+            )
+            if self._max_retries > 0:
+                # Don't raise RetriesExhaustedError unless retries are actually attempted
```
Cito, line 329: I think we should add some kind of exponential backoff here instead of retrying immediately. This should also be made configurable.

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -262,3 +429,154 @@ async def run(self, forever: bool = True) -> None:
...
+                topic=self._retry_topic,
+                key=event.key,
+            )
+            logging.info(
+                "Published an event with type '%s' to the retry topic '%s'",
+                type_,
+                self._retry_topic,
+            )
+
+    async def run(self, ignore: bool = False) -> None:
```
Cito, line 566: Instead of passing a flag, we could pass a callback function that can inspect the event and then ignore or pass it - maybe even allow this callback to modify the payload when passing it to the retry topic.

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -177,76 +210,210 @@ async def construct(
...
+        if topic == self._retry_topic:
+            topic = str(event.value.get(ORIGINAL_TOPIC_FIELD, ""))
+            logging.info(
+                "Received previously failed event from topic '%s' for retry.", topic
+            )
+        return topic
+
+    async def _publish_to_dlq(self, *, event: ExtractedEventInfo):
+        """Publish the event to the DLQ topic."""
+        dlq_payload = {**event.payload, ORIGINAL_TOPIC_FIELD: event.topic}
```
Cito, line 269: Wouldn't it make more sense to put the original topic into the header instead of the payload? I believe it is intended to be used for this kind of metadata.

Cito (changes requested):
Awesome work.

I have left some feedback below.

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -177,76 +210,210 @@ async def construct(
...
             translator (EventSubscriberProtocol):
                 The translator that translates between the protocol (mentioned in the
                 type annotation) and an application-specific port
                 (according to the triple hexagonal architecture).
+            publisher:
+                running instance of publishing provider that implements the
+                EventPublisherProtocol, such as KafkaEventPublisher. Can be None if
+                not using the dead letter queue.
+            config:
+                the KafkaRetryConfig instance containing dlq topics and retry allowance.
```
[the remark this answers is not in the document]
TheByronHimes: This was from a previous iteration and I apparently never caught it afterward. And thanks for the standardization vote

src/hexkit/providers/akafka/provider/eventsub.py
TheByronHimes: That's what I originally wanted to do but felt like the protocol was too holy to change

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -262,3 +429,154 @@ async def run(self, forever: bool = True) -> None:
...
+            super().__init__(msg)
+
+    @classmethod
+    @asynccontextmanager
+    async def construct(
+        cls,
+        *,
+        config: KafkaConfig,
+        kafka_consumer_cls: type[KafkaConsumerCompatible] = AIOKafkaConsumer,
+        publisher: EventPublisherProtocol,
```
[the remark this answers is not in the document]
TheByronHimes: Changed in [2c9a144](https://github.com/ghga-de/hexkit/pull/123/commits/2c9a14484ad3b5460a0daa21d76347dbedd5279f)

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -177,76 +210,210 @@ async def construct(
...
         """Please do not call directly! Should be called by the `construct` method.
         Args:
             consumer:
-                hands over a started AIOKafkaProducer.
+                hands over a started AIOKafkaConsumer.
             translator (EventSubscriberProtocol):
                 The translator that translates between the protocol (mentioned in the
                 type annotation) and an application-specific port
                 (according to the triple hexagonal architecture).
+            publisher:
```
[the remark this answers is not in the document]
TheByronHimes: Changed in [2c9a144](https://github.com/ghga-de/hexkit/pull/123/commits/2c9a14484ad3b5460a0daa21d76347dbedd5279f)

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -154,6 +176,10 @@ async def construct(
...
```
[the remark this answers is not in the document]
TheByronHimes: Changed in [2c9a144](https://github.com/ghga-de/hexkit/pull/123/commits/2c9a14484ad3b5460a0daa21d76347dbedd5279f)

src/hexkit/providers/akafka/provider/daosub.py
```diff
@@ -110,21 +112,34 @@ async def construct(
         *,
         config: KafkaConfig,
         translators: Sequence[DaoSubscriberProtocol],
+        publisher: Optional[EventPublisherProtocol] = None,
```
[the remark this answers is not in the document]
TheByronHimes: See [2c9a144](https://github.com/ghga-de/hexkit/pull/123/commits/2c9a14484ad3b5460a0daa21d76347dbedd5279f)

src/hexkit/providers/akafka/config.py
```diff
@@ -76,3 +76,49 @@ class KafkaConfig(BaseSettings):
             + " newly correlation ID will be generated and used in the event header."
         ),
     )
+    kafka_dlq_topic: str = Field(
+        default="",
+        description="The name of the service-specific topic used for the dead letter queue.",
+        examples=["dcs-dlq", "ifrs-dlq", "mass-dlq"],
```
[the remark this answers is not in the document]
TheByronHimes: I didn't know about that parameter. See [210e3a9](https://github.com/ghga-de/hexkit/pull/123/commits/210e3a92850c49287c3bdc1dc20794948085a33d)

src/hexkit/providers/akafka/provider/eventsub.py
Cito: The payload is even holier 😇

src/hexkit/providers/akafka/provider/eventsub.py
TheByronHimes: Agreed. Added in [fcc9a71](https://github.com/ghga-de/hexkit/pull/123/commits/fcc9a71f40620c5a9fcb80fc17385ff9d2c03208)

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -177,76 +210,210 @@ async def construct(
...
+                # Don't raise RetriesExhaustedError unless retries are actually attempted
+                try:
+                    await self._retry_event(event=event, retries_left=self._max_retries)
+                except self.RetriesExhaustedError as retry_error:
+                    # Publish to the DLQ or re-raise the exception
+                    logging.warning(retry_error)
+                    if self._enable_dlq:
+                        await self._publish_to_dlq(event=event)
+                    else:
+                        raise retry_error
```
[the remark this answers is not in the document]
TheByronHimes: I refactored this section in [791515b](https://github.com/ghga-de/hexkit/pull/123/commits/791515b12b4df1f7c0cb94fbfb17da72aaf347c4)

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -177,76 +210,210 @@ async def construct(
...
+            payload=dlq_payload,
+            type_=event.type_,
+            topic=self._dlq_topic,
+            key=event.key,
         )
+        logging.info("Published event to DLQ topic '%s'", self._dlq_topic)
+
+    async def _retry_event(self, *, event: ExtractedEventInfo, retries_left: int):
+        """Retry the event until the maximum number of retries is reached."""
+        retries_left -= 1
```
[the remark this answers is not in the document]
TheByronHimes: Good idea. Added in [791515b](https://github.com/ghga-de/hexkit/pull/123/commits/791515b12b4df1f7c0cb94fbfb17da72aaf347c4)

src/hexkit/providers/akafka/provider/eventsub.py
TheByronHimes: I ran with that and modified the event pub protocol in [ee00b5e](https://github.com/ghga-de/hexkit/pull/123/commits/ee00b5e224ef53790692f82b54f41e0d1b1d77ba). The consequence is that this is a breaking change, so the branch was rebased to v4 and the PR retargeted.

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -157,6 +191,10 @@ async def construct(
                 The translator that translates between the protocol (mentioned in the
                 type annotation) and an application-specific port
                 (according to the triple hexagonal architecture).
+            dlq_publisher:
+                running instance of publishing provider that implements the
```
Cito, line 195:
```suggestion
                A running instance of a publishing provider that implements the
```

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -181,76 +226,220 @@ async def construct(
...
         Args:
             consumer:
-                hands over a started AIOKafkaProducer.
+                hands over a started AIOKafkaConsumer.
             translator (EventSubscriberProtocol):
                 The translator that translates between the protocol (mentioned in the
                 type annotation) and an application-specific port
                 (according to the triple hexagonal architecture).
+            dlq_publisher:
+                running instance of publishing provider that implements the
```
Cito, line 258:
```suggestion
                A running instance of a publishing provider that implements the
```

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -181,76 +226,220 @@ async def construct(
...
+                    raise  # re-raise Exception
+                await self._publish_to_dlq(event=event)
+                return
+
+            # Don't raise RetriesExhaustedError unless retries are actually attempted
+            try:
+                await self._retry_event(event=event, retries_left=self._max_retries)
+            except (self.RetriesExhaustedError, ValueError) as retry_error:
+                # In case of ValueError, proceed with flow because that error is
+                # likely due to a hexkit bug, not a config/event error.
```
Cito, line 369: This sounds a bit hacky. ValueError could also be produced by the translator or the core.

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -181,76 +226,220 @@ async def construct(
...
+
+        If the retry fails, this method is called again with `retries_left` decremented.
+
+        Raises:
+        - `RetriesExhaustedError`: If all retries are exhausted without success.
+        - `ValueError`: If `retries_left` is invalid.
+        """
+        # Check if retries_left is valid.
+        if not 0 < retries_left <= self._max_retries:
+            error = ValueError(
```
Cito, line 297: Maybe also raise a RetriesExhaustedError in this case since it probably should be dealt with in the same way.

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -181,76 +226,220 @@ async def construct(
...
+                event_info.topic,
+            )
+        return event_info
+
+    def _validate_extracted_info(self, event: ExtractedEventInfo):
+        """Extract and validate the event, returning the correlation ID and the extracted info."""
+        correlation_id = event.headers.get("correlation_id", "")
+        errors = []
+        if not event.type_:
+            errors.append("type_ is empty")
```
Cito, line 392:
```suggestion
            errors.append("event type is empty")
```

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -181,76 +226,220 @@ async def construct(
...
+        elif event.type_ not in self._types_whitelist:
+            errors.append(f"type_ '{event.type_}' is not in the whitelist")
```
Cito, line 394:
```suggestion
            errors.append(f"event type '{event.type_}' is not in the whitelist")
```

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -266,3 +455,233 @@ async def run(self, forever: bool = True) -> None:
...
+def validate_dlq_headers(event: ConsumerEvent) -> None:
+    """Validate the headers that should be populated on every DLQ event.
+
+    Raises:
+    - `DLQValidationError`: If any headers are determined to be invalid.
+    """
+    headers = headers_as_dict(event)
+    expected_headers = ["type", "correlation_id", ORIGINAL_TOPIC_FIELD]
+    invalid_headers = [
+        key for key in expected_headers if key not in headers or not headers[key]
```
Cito, line 489:
```suggestion
        key for key in expected_headers if not headers.get(key)
```

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -266,3 +455,233 @@ async def run(self, forever: bool = True) -> None:
...
+
+        Args:
+        - `config`:
+            Config parameters needed for connecting to Apache Kafka.
+        - `dlq_publisher`:
+            running instance of publishing provider that implements the
+            EventPublisherProtocol, such as KafkaEventPublisher. It is used to publish
+            events to the configured retry topic.
+        - `kafka_consumer_cls`:
+            Overwrite the used Kafka consumer class . Only intended for unit testing.
```
Cito, line 544:
```suggestion
            Overwrite the used Kafka consumer class. Only intended for unit testing.
```

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -266,3 +455,233 @@ async def run(self, forever: bool = True) -> None:
...
+            enable_auto_commit=False,
+            key_deserializer=lambda event_key: event_key.decode("ascii"),
+            value_deserializer=lambda event_value: json.loads(
+                event_value.decode("ascii")
+            ),
+            max_partition_fetch_bytes=config.kafka_max_message_size,
+        )
+
+        try:
+            await consumer.start()
```
Cito, line 574:
```suggestion
        await consumer.start()
        try:
```

Cito: If the consumer was not started, it should not be closed.

Cito (approved):
Sorry it took so long to review.

Just some minor suggestions.

src/hexkit/providers/akafka/provider/eventsub.py
TheByronHimes: Fixed in [4d2b5c9](https://github.com/ghga-de/hexkit/pull/123/commits/4d2b5c920f4a048d97cb147a5a3e772af03abee8)

src/hexkit/providers/akafka/provider/eventsub.py
TheByronHimes: This and the similar comment above fixed in [e6271fc](https://github.com/ghga-de/hexkit/pull/123/commits/e6271fca629f1a977f3608e7d8554fd3e30e33cb)

src/hexkit/providers/akafka/provider/eventsub.py
TheByronHimes: [e6271fc](https://github.com/ghga-de/hexkit/pull/123/commits/e6271fca629f1a977f3608e7d8554fd3e30e33cb)

src/hexkit/providers/akafka/provider/eventsub.py
TheByronHimes: Good catch: [792763c](https://github.com/ghga-de/hexkit/pull/123/commits/792763caf0ffe31c4d050984d3c4c38085390fa8)

src/hexkit/providers/akafka/provider/eventsub.py
TheByronHimes: If the ValueError was raised in the core or translator, it would have been caught in [in the retry method](https://github.com/ghga-de/hexkit/blob/b5ac74e1ace3cbb1414be72cea92bf41b06e1efc/src/hexkit/providers/akafka/provider/eventsub.py#L326) and ultimately re-raised as a RetriesExhaustedError. Maybe the comment should be removed or changed for clarity.

src/hexkit/providers/akafka/provider/eventsub.py
TheByronHimes: The `RetriesExhaustedError` represents a specific outcome -- that an event failed, was retried at least once, and failed for all of the allowed retries. The current logic should already prevent the function from being called with `retries_left` outside of the above criteria. If it somehow _does_ get called with an invalid value, `ValueError` is a much better fit to convey what happened than `RetriesExhaustedError`, even if the downstream handling is identical. Between logs and the traceback, I think this will provide the most clarity for diagnostics.

Cito: Yes, it should never happen. I agree that it would not be really clean to reuse this error. Maybe raise a specific subclass of ValueError that you can catch to make it cleaner.

TheByronHimes: Alright, I used a new error class in [27edd4a](https://github.com/ghga-de/hexkit/pull/123/commits/27edd4a45f1a0b25d7d66b8e8d008a1d576c7d1d), which also updates the comment mentioned above

src/hexkit/providers/akafka/provider/eventsub.py
TheByronHimes: See [27edd4a](https://github.com/ghga-de/hexkit/pull/123/commits/27edd4a45f1a0b25d7d66b8e8d008a1d576c7d1d)

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -266,3 +464,231 @@ async def run(self, forever: bool = True) -> None:
...
+
+def validate_dlq_headers(event: ConsumerEvent) -> None:
+    """Validate the headers that should be populated on every DLQ event.
+
+    Raises:
+    - `DLQValidationError`: If any headers are determined to be invalid.
+    """
+    headers = headers_as_dict(event)
+    expected_headers = ["type", "correlation_id", ORIGINAL_TOPIC_FIELD]
+    invalid_headers = [key for key in expected_headers if not headers.get(key, None)]
```
Cito, line 497:
```suggestion
    invalid_headers = [key for key in expected_headers if not headers.get(key)]
```

None is already the "default default".

Cito (approved):

src/hexkit/providers/akafka/provider/eventsub.py
Cito: A better name would be `missing_headers` instead of `invalid_headers`.

mephenor (approved):
Very nice.
I encountered only some minor inconsistencies in the documentation, but nothing in the public facing part.

tests/unit/test_dlqsub.py
```diff
@@ -0,0 +1,708 @@
...
+        assert parsed_log.startswith("Ignored event of type 'test_type': retry")
+        assert parsed_log.endswith("errors: topic is empty")
+
+
+@pytest.mark.asyncio()
+async def test_dlq_subscriber_ignore(kafka: KafkaFixture, caplog_debug):
+    """Test what happens when a DLQ Subscriber is instructed to ignore an event."""
+    config = make_config(kafka.config)
+
+    # make an event without the original_topic field in the payload
```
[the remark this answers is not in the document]
TheByronHimes: Nice catch, this test should have been updated when I moved the original topic from the payload to the header, but clearly I missed it.

tests/unit/test_dlqsub.py
```diff
@@ -0,0 +1,708 @@
...
+    translator = FailSwitchTranslator(
+        topics_of_interest=["test-topic"], types_of_interest=["test_type"], fail=True
+    )
+    async with KafkaEventSubscriber.construct(
+        config=config, translator=translator, dlq_publisher=dummy_publisher
+    ) as retry_sub:
+        with pytest.raises(RuntimeError) if not enable_dlq else nullcontext():
+            await retry_sub.run(forever=False)
+
+    # Verify that the event was retried twice after initial failure
```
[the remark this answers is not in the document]
TheByronHimes: Yes, I updated the wording to be more accurate

tests/unit/test_dlqsub.py
```diff
@@ -0,0 +1,708 @@
...
+    deletions: list[str]
+
+    def __init__(self, *, fail: bool = False) -> None:
+        self.upsertions = []
+        self.deletions = []
+        self.fail = fail
+
+    async def changed(self, resource_id: str, update: OutboxDto) -> None:
+        """Dummy"""
+        if ORIGINAL_TOPIC_FIELD in update.model_dump():
```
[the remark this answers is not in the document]
TheByronHimes: Ah, thank you. Another vestige of the payload -> header move

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -266,3 +467,231 @@ async def run(self, forever: bool = True) -> None:
...
+
+
+# Define function signature for the DLQ processor.
+DLQEventProcessor = Callable[[ConsumerEvent], Awaitable[Optional[ExtractedEventInfo]]]
+
+
+class DLQValidationError(RuntimeError):
+    """Raised when an event from the DLQ fails validation."""
+
+    def __init__(self, *, event: ConsumerEvent, error: str):
```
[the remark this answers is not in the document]
TheByronHimes: Changed it to `reason`

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -266,3 +467,231 @@ async def run(self, forever: bool = True) -> None:
...
+        msg = f"DLQ Event '{get_event_label(event)}' is invalid: {error}"
+        super().__init__(msg)
+
+
+class DLQProcessingError(RuntimeError):
+    """Raised when an error occurs while processing an event from the DLQ."""
+
+    def __init__(self, *, event: ConsumerEvent, error: str):
```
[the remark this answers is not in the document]
TheByronHimes: See above

mephenor (approved):

PR #125, opened 2024-08-27
Handle Outbox Documents Lacking Metadata (GSI-924)

TheByronHimes:
Before this PR, updating or deleting existing documents with an outbox DAO would result in a `ResourceNotFoundError` because the update/delete query assumed the existence of the `__metadata__` field. This PR modifies the two relevant methods so the query results additionally include documents with a matching ID but no `__metadata__` field.

Bumps version from `3.5.0` to `3.5.1`.


mephenor (approved):
LGTM

PR #126, opened 2024-08-28
Federated s3 fixture (GSI-965)

TheByronHimes:
Adds the `S3MultiContainerFixture` and `FederatedS3Fixture` that were introduced in the SMS. Maintains the same persistent and clean fixture variations that exist with other fixtures. 
Because the `S3MultiContainerFixture` must know ahead of time about the containers it needs to create, it expects a fixture named `storage_aliases` to be defined with the same scope. If you try to use the fixture without having defined the `storage_aliases` fixture, you'll get an error.

Bumps version to `3.6.0`.


tests/unit/test_federated_s3.py
```diff
@@ -0,0 +1,84 @@
...
+@pytest.fixture(scope="session")
+def storage_aliases():
+    """Return the storage aliases for the federated S3 fixture."""
+    return STORAGE_ALIASES
+
+
+async def test_get_configs_by_alias(federated_s3: FederatedS3Fixture):
+    """Test that `get_configs_by_alias` returns the configs of each storage by name."""
+    configs = federated_s3.get_configs_by_alias()
+    assert configs.keys() == set(STORAGE_ALIASES)
```
Cito, line 44:
```suggestion
    assert set(configs) == set(STORAGE_ALIASES)
```

Since it's not obvious that you can compare a view object with a set.

tests/unit/test_federated_s3.py
```diff
@@ -0,0 +1,84 @@
...
+    assert not await federated_s3.storages[
+        SECONDARY_STORAGE_ALIAS
+    ].storage.does_bucket_exist(bucket_id="empty")
+
+
+async def test_multi_container_fixture(
+    s3_multi_container: S3MultiContainerFixture,
+):
+    """Test that the multi container fixture actually uses separate S3 instances."""
+    assert s3_multi_container.s3_containers.keys() == set(STORAGE_ALIASES)
```
Cito, line 81:
```suggestion
    assert set(s3_multi_container.s3_containers) == set(STORAGE_ALIASES)
```

see above

Cito (approved):
Nice. Just a minor suggestion.

Cito (approved):

PR #127, opened 2024-09-02
Simplify DAO API (GSI-789)

TheByronHimes:
Replaces the `DaoCommons`,  `DaoNaturalId`, and `DaoSurrogateId` protocol classes with a single `Dao` protocol. 

Replaces the parallel MongoDb DAO classes with a single `MongoDbDao` class.

Removes the `uuid4_id_generator` since it's essentially replaced by the provided `BaseModelWithId` class.


tests/integration/test_mongodb.py
```diff
@@ -228,16 +145,14 @@ async def test_dao_insert_natural_id_happy(mongodb: MongoDbFixture):
 
 
 async def test_dao_upsert_natural_id_happy(mongodb: MongoDbFixture):
```
Cito, line 147:
```suggestion
async def test_dao_upsert_happy(mongodb: MongoDbFixture):
```

tests/integration/test_mongodb.py
```diff
@@ -140,72 +121,8 @@ async def test_empty_collections(mongodb: MongoDbFixture):
...
-    await dao.delete(resource_inserted.id)
-
-    # confirm that the resource was deleted:
-    with pytest.raises(ResourceNotFoundError):
-        _ = await dao.get_by_id(resource_inserted.id)
-
-
-async def test_dao_insert_natural_id_happy(mongodb: MongoDbFixture):
-    """Tests the happy path of inserting a new resource in a natural ID setting."""
+async def test_dao_insert_happy(mongodb: MongoDbFixture):
```
Cito, line 124: I think we also need to test the `insert_unhappy` behavior (e.g. when the id already exists).

Cito: I would add the happy CRUD test as well even when the individual steps are already tested.

The CRUD test is also useful as a compact documentation how to use the DAO.

tests/unit/test_dao.py
```diff
@@ -16,87 +16,40 @@
 
 """Testing the dao factory protocol."""
```
Cito, line 17:
```suggestion
"""Testing the DAO factory protocol."""
```

Cito (changes requested):
Thanks - this looks much cleaner and simpler now, nice.

See some suggestions below that we should discuss.

tests/integration/test_mongodb.py
TheByronHimes: The duplicate ID behavior is already tested at the end of that test. After the successful insertion, it attempts a repeat and looks for the error. I will move it into a separate test though.

Cito: Yes, either remove the "happy" from the name or (better) move it to a separate test.

TheByronHimes: Please see [8036561](https://github.com/ghga-de/hexkit/pull/127/commits/80365614ba7c94bec114837add7cdb9bcf901c06) for the update

Cito: I wonder if we should add a similar, but separate test where we mock uuid.hex to produce duplicate values (like 1, 1, 2, 2 , ...) to simulate the (extremely unlikely) case that a duplicate UUID is produced. The test could then also document what happens in this case and how it can be recovered (simple retry). Maybe the insert method could have a parameter `retries_if_duplicate`?

src/hexkit/protocols/dao.py
```diff
@@ -99,11 +95,34 @@ def __init__(self, *, mapping: Mapping[str, str]):
...
+class BaseModelWithId(BaseModel):
+    """A convenience base class for DTOs that have an ID field and require automatic ID
+    generation. When the model is created, a new UUID4 is generated and assigned to the
+    ID field.
     """
 
+    id: str = Field(
+        default_factory=lambda: str(uuid4()),
+        description="The unique identifier of the resource.",
+    )
```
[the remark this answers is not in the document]
TheByronHimes: Pydantic already offers [`UUID1`, `UUID3`, `UUID4`, and `UUID5` types](https://docs.pydantic.dev/latest/api/types/#pydantic.types.UUID1). Does that not cover it sufficiently enough? That provides validation and typing, and the `default_factory` parameter is straightforward.

Given that the description and name often need to be more specific (which I didn't think about) should we just remove the `BaseModelWithId`?

tests/integration/test_mongodb.py
TheByronHimes: Considering ID generation happens before/outside of `insert()`, I feel the responsibility of the interpreting and handling this case falls on the user. I don't think there's a way for a UUID collision to be detected by a the DAO, unless we're talking about different things.

src/hexkit/protocols/dao.py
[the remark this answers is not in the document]
Cito: You're right, we should make use of these. However, they are just types without default factory.

How about this?

In hexkit:

```python
from functools import partial
from uuid import uuid4

from pydantic import UUID4, BaseModel, Field

UUID4Field = partial(Field, default_factory=uuid4)
```

In user code:

```python

from hexkit import UUID4, UUID4Field

class User(BaseModel):
    """A user of our system."""
    id: UUID4 = UUID4Field(
        title="User ID", description="The unique identifier for the user."
    )
   name: str = ...
```

Just to make using this a bit more convenient.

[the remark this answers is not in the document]
Cito: The UUID4 fields should be (de)serialized properly in Mongo without further ado on our side I think.

tests/integration/test_mongodb.py
Cito: That's true. The insert method cannot just retry, but would need to do something like this to recreate the id before trying to insert again;

```python
dto = Dto(**dto.model_dump(exclude={self._id_field}))
```

Cito: But the problem is then that the user would not be aware that the id has changed since the dto is not passed back.

Cito: Probably this is too much magic, and we should leave the responsibility for handling this to the user as you said.

But it would still be good to have a test as example how to do handle this case.

src/hexkit/protocols/dao.py
```diff
@@ -95,16 +96,13 @@ def __init__(self, *, mapping: Mapping[str, str]):
...
+    """Generate a UUID4 string."""
+    return str(uuid4())
+
 
-    id: str = Field(
-        default_factory=lambda: str(uuid4()),
-        description="The unique identifier of the resource.",
-    )
+# provide standardized default factory for UUID4 fields
+UUID4Field = partial(Field, default_factory=get_uuid4_str)
```
Cito, line 105: Is there a reason why you introduce the `get_uuid4_str` function? It should work with

```python
UUID4Field = partial(Field, default_factory=uuid4)
```

Even if the Field uses a `str` type instead of the `UUID4` type.

tests/integration/test_mongodb.py
```diff
@@ -39,10 +39,14 @@
...
-class ExampleDto(BaseModelWithId):
-    """Example DTO with an auto-generated ID."""
+class ExampleDto(BaseModel):
+    """Example DTO with an auto-generated ID.
+
+    Does not use `UUID4` type in order to facilitate testing.
+    """
 
     model_config = ConfigDict(frozen=True)
+    id: str = UUID4Field(description="The ID of the resource.")
```
Cito, line 49: The tests should reflect the real/recommended usage of the provided functionality, both in order to test real world usage and since they serve as a "living documentation" on how to use the functionality.

Therefore we should really run the tests with a `UUID4` type here. Unless we recommend that users should also use the `str` type. Which I think they shouldn't, since the UUID4 validation provides additional safety that invalid values can never make their way into the database. (This was actually one of the attack vectors in the penetration tests!)

src/hexkit/protocols/dao.py
```diff
@@ -38,7 +39,7 @@
     "MultipleHitsFoundError",
     "Dao",
     "DaoFactoryProtocol",
-    "BaseModelWithId",
+    "UUID4Field",
```
Cito, line 42: Maybe also import and re-export UUID4 for convenience, so that both can be imported in the same statement.

tests/integration/test_mongodb.py
Cito: I would still add a separate test case that shows what happens in the case where UUID4Field creates a duplicate value (can be simulated by mocking uuid4) and how the user could handle this case (by recreating the object excluding the id field and retrying). Again, mainly meant as living documentation.

Cito (changes requested):
Some more questions/suggestions.

src/hexkit/protocols/dao.py
TheByronHimes: BSON/Native UUIDs didn't work right and we've used `str` for virtually all IDs so far, so it made sense to me to continue that way. After discussing with you offline, we decided to look into a comprehensive switch from string (JSON) format in the DB to BSON as another ticket.  
For the `UUID4Field`, I've reverted to the uuid4 function.

tests/integration/test_mongodb.py
TheByronHimes: Good point. The tests have been updated to use UUID4

src/hexkit/protocols/dao.py
TheByronHimes: In most cases where you would need `UUID4Field`, you're defining a model that subclasses `BaseModel`. If you're importing `BaseModel`, you can also import `UUID4`, so it probably doesn't offer much of an advantage to import-export it here.

Cito: Agreed.

Cito (approved):

PR #128, opened 2024-09-03
Add DAO protocol deprecation warnings (GSI-976)

TheByronHimes:


Cito (changes requested):
We should also add deprecation warnings in the init methods of the MongoDbDaoNaturalId and MongoDbDaoSurrogateId, telling the user how to update - in the first case, just use MobgoDbDao, in the second case, use MobgoDbDao, remove the creation model, and replace the normal id Field with an UUID4Field in the DTO class.

src/hexkit/providers/mongodb/provider.py
```diff
@@ -163,6 +164,17 @@ def __init__(
         self._document_to_dto = document_to_dto
         self._dto_to_document = dto_to_document
 
+        warnings.warn(
+            "The MongoDbDaoBase class and subclasses are deprecated as of v3.6 and will"
+            + " be replaced in hexkit v4 with the new MongoDbDao class. To replace the"
+            + " MongoDbDaoNaturalId: just use the new class. To replace the"
+            + " MongoDbDaoSurrogateId: use the new class, remove the creation model and"
+            + " id_generator parameters, and make sure the supplied DTO model includes"
+            + " the ID field.",
```
Cito, line 173:
```suggestion
            + " the ID field with a default factory for generating new IDs.",
```

Cito (approved):
Looks good. One minor addition below.

PR #132, opened 2024-09-13
Configurable MongoDB Timeout (GSI-987)

TheByronHimes:
This adds `mongo_timeout` to the `MongoDbConfig` class. The value is in seconds. The default is `None`, which will preserve the MongoDb default behavior. I chose this opt-in approach because the appropriate value varies by use-case and I don't know that we have a sense for what it should be, if set at all.

**More info:**  
The timeout config value is converted from seconds to milliseconds when instantiating the -DaoFactory and supplied to the client.  
The timeout is then applied to all subsequent API requests through that client (effectively the DAO). 

When a timeout occurs in Pymongo, it will result in one of several errors that subclass `PyMongoError`. PyMongoError is also the base class for other error types, and the timeout-caused errors can be identified by `error.timeout`, a property that defaults to `False`.  
DB operations in the MongoDb and MongoKafka providers are wrapped with an error translation function that reraises the various flavors of timeout error as `DbTimeoutError`.

**Testing:**  
The tests added are simple. There is a test for the translation wrapper itself, then a test (added for both Mongo/MongoKafka) that adds the shortest configurable timeout and attempts each of the DAO functions to ensure the timeouts are caught. The purpose of the test is not to test translation but to make sure that error handling (via the wrapper) is applied everywhere it should be.

### DIFF
--- a/src/hexkit/protocols/dao.py
+++ b/src/hexkit/protocols/dao.py
@@ -47,7 +47,11 @@ __all__ = [
 Dto = TypeVar("Dto", bound=BaseModel)
 
 
-class ResourceNotFoundError(RuntimeError):
+class DaoError(RuntimeError):
+    """Base for all errors related to DAO operations."""
+
+
+class ResourceNotFoundError(DaoError):
     """Raised when a requested resource did not exist."""
 
     def __init__(self, *, id_: ID):
@@ -55,7 +59,7 @@ class ResourceNotFoundError(RuntimeError):
         super().__init__(message)
 
 
-class ResourceAlreadyExistsError(RuntimeError):
+class ResourceAlreadyExistsError(DaoError):
     """Raised when a resource did unexpectedly exist."""
 
     def __init__(self, *, id_: ID):
@@ -63,7 +67,7 @@ class ResourceAlreadyExistsError(RuntimeError):
         super().__init__(message)
 
 
-class FindError(RuntimeError):
+class FindError(DaoError):
     """Base for all error related to DAO find operations."""
 
 
@@ -97,11 +101,11 @@ class NoHitsFoundError(FindError):
         super().__init__(message)
 
 
-class DbTimeoutError(RuntimeError):
+class DbTimeoutError(DaoError):
     """Raised when a database operation timed out."""
 
-    def __init__(self, reason: str):
-        message = f"The database operation timed out: {reason}"
+    def __init__(self, details: str):
+        message = f"The database operation timed out: {details}"
         super().__init__(message)
 
 
--- a/src/hexkit/protocols/dao.py
+++ b/src/hexkit/protocols/dao.py
@@ -97,6 +97,14 @@ class NoHitsFoundError(FindError):
         super().__init__(message)
 
 
+class DbTimeoutError(RuntimeError):
+    """Raised when a database operation timed out."""
+
+    def __init__(self, reason: str):
+        message = f"The database operation timed out: {reason}"
+        super().__init__(message)
+
+
 # provide standardized default factory for UUID4 fields
 UUID4Field = partial(Field, default_factory=uuid4)
 
--- a/src/hexkit/providers/mongodb/provider.py
+++ b/src/hexkit/providers/mongodb/provider.py
@@ -23,7 +23,7 @@ Utilities for testing are located in `./testutils.py`.
 
 import json
 from collections.abc import AsyncIterator, Collection, Mapping
-from contextlib import AbstractAsyncContextManager
+from contextlib import AbstractAsyncContextManager, contextmanager
 from datetime import date, datetime
 from functools import partial
 from pathlib import Path
@@ -34,12 +34,13 @@ from motor.core import AgnosticCollection
 from motor.motor_asyncio import AsyncIOMotorClient
 from pydantic import Field, MongoDsn, PositiveInt, Secret
 from pydantic_settings import BaseSettings
-from pymongo.errors import DuplicateKeyError
+from pymongo.errors import DuplicateKeyError, PyMongoError
 
 from hexkit.custom_types import ID
 from hexkit.protocols.dao import (
     Dao,
     DaoFactoryProtocol,
+    DbTimeoutError,
     Dto,
     InvalidFindMappingError,
     MultipleHitsFoundError,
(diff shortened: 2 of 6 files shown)


src/hexkit/providers/mongodb/provider.py
```diff
@@ -49,7 +50,23 @@
...
+
+@contextmanager
+def translate_timeout_error():
+    """Catch PyMongoError and re-raise it as DbTimeoutError if it is a timeout error."""
+    try:
+        yield
+    except PyMongoError as exc:
+        if exc.timeout:  # denotes timeout-related errors in pymongo
+            raise DbTimeoutError(str(exc)) from exc
+        raise
```
Cito, line 69: We might reraise this as `raise DbError(str(exc)) from exc` to make it more consistent. Where `DbError` would be derived from `RuntimeError` or `DaoError`.

src/hexkit/protocols/dao.py
```diff
@@ -97,6 +97,14 @@ def __init__(self, *, mapping: Mapping[str, str]):
         super().__init__(message)
 
 
+class DbTimeoutError(RuntimeError):
+    """Raised when a database operation timed out."""
+
+    def __init__(self, reason: str):
```
Cito, line 103: Maybe something like `details` or `raw_message` instead of `reason`? Because the reason for the error is the timeout and the reason for the timeout is probably not specified in this message.

src/hexkit/providers/mongodb/provider.py
```diff
@@ -380,6 +404,17 @@ class MongoDbConfig(BaseSettings):
         examples=["my-database"],
         description="Name of the database located on the MongoDB server.",
     )
+    mongo_timeout: Union[PositiveInt, None] = Field(
+        default=None,
+        examples=[300, 600, None],
+        description=(
+            "Timeout in seconds for API calls to MongoDB. The timeout applies all steps"
```
Cito, line 411:
```suggestion
            "Timeout in seconds for API calls to MongoDB. The timeout applies to all steps"
```

src/hexkit/providers/mongodb/provider.py
```diff
@@ -396,10 +431,16 @@ def __init__(
...
+            int(config.mongo_timeout * 1000)
+            if config.mongo_timeout is not None
+            else None
+        )
 
         # get a database-specific client:
         self._client: AsyncIOMotorClient = AsyncIOMotorClient(
-            str(self._config.mongo_dsn.get_secret_value())
+            str(self._config.mongo_dsn.get_secret_value()),
+            timeoutms=timeout_ms,
```
Cito, line 443: Shouldn't this be `timeoutMS` - see [docs](https://pymongo.readthedocs.io/en/stable/api/pymongo/mongo_client.html)? Why is this working though?

tests/unit/test_mongokafka.py
```diff
@@ -15,8 +15,15 @@
...
+import pytest
+from pydantic import BaseModel
+
+from hexkit.correlation import set_new_correlation_id
+from hexkit.protocols.dao import DbTimeoutError
 from hexkit.providers.mongokafka import MongoKafkaConfig
+from hexkit.providers.mongokafka.provider import MongoKafkaDaoPublisherFactory
 
 
 def make_mongokafka_config(kafka_max_message_size: int = 1048576) -> MongoKafkaConfig:
```
Cito, line 29: You should mention here that the mongo DSN is supposed to *not* be running.

Cito (changes requested):
Looks good, but found some things you should look at again.

src/hexkit/providers/mongodb/provider.py
TheByronHimes: They parse the kwargs with a case-insensitive dictionary

src/hexkit/providers/mongodb/provider.py
TheByronHimes: I think a `DaoError` adds enough for now. We can add more layers in the future if we really need to.

src/hexkit/protocols/dao.py
```diff
@@ -97,6 +97,14 @@ def __init__(self, *, mapping: Mapping[str, str]):
         super().__init__(message)
 
 
+class DbTimeoutError(RuntimeError):
```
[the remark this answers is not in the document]
TheByronHimes: Sure, that seems good.

src/hexkit/providers/mongodb/provider.py
Cito: Probably to be compatible with mongo clients in other languages. I think we should use the documented spelling though.

tests/unit/test_mongokafka.py
Cito: To the `make_mongokafka_config` function.

src/hexkit/providers/mongodb/provider.py
Cito: Ok. The main thing is that we make it possible to catch any DAO related errors with a single exception statement.

src/hexkit/providers/mongodb/provider.py
TheByronHimes: Reraising non-timeout errors here as anything else, like DaoError, will obfuscate the actual error. The other error translation that occurs is the `DuplicateKeyError` -> `ResourceAlreadyExistsError` in the insertion method, and that requires the resource ID. So a simple `raise` is probably the way to go until we have a need for more complex error translation.

Cito: Right, since we need to differentiate *after* this translation.

The actual problem I want to solve is that the code using the DAO has a way to catch all possible database exceptions without knowing which kind of database is used (since the DAO is working on a more abstract level). Maybe you have a better idea?

Cito: I mean there could be other database errors that are neither timeout errors nor one of the errors that are caught and re-re-raised by the provider (currently only DuplicateKeyError).

tests/unit/test_mongokafka.py
TheByronHimes: Oh I see now. The way it captured the import addition threw me off. See [89c5d87](https://github.com/ghga-de/hexkit/pull/132/commits/89c5d870b72b8c52bbba90e96a5069c8cb2e877f) for the docstring update

src/hexkit/protocols/dao.py
TheByronHimes: Changed to `details` here: [89c5d87](https://github.com/ghga-de/hexkit/pull/132/commits/89c5d870b72b8c52bbba90e96a5069c8cb2e877f)

src/hexkit/protocols/dao.py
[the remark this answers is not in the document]
TheByronHimes: Added in [89c5d87](https://github.com/ghga-de/hexkit/pull/132/commits/89c5d870b72b8c52bbba90e96a5069c8cb2e877f)

src/hexkit/providers/mongodb/provider.py
TheByronHimes: I see what you mean. Then a translation function (maybe a function decorator if it works) can be used to catch `PyMongoError` (after the given function has completed) and re-raise it as the `DaoError`. That should work because any other error translation would have already taken place, so the only PyMongoErrors that go through would be the ones we _want_ to translate to the generic DaoError class. It will also preserve non-DB errors that occur in the flow.

src/hexkit/providers/mongodb/provider.py
Cito: I know it's actually not in the scope of this PR but since we are already working on error handling, we may want to improve this as well.

TheByronHimes: I will move this extra error translation into a subsequent PR.

Cito (approved):

PR #135, opened 2024-11-04
Include exception info in DLQ event headers (GSI-1119)

TheByronHimes:
DLQ events should automatically contain the underlying exception information. This does not bump the version number because it should really be part of the 4.0 release.

This PR modifies the process so that `KafkaEventSubscriber._publish_to_dlq()` now requires an exception instance.
The exception instance is translated into 2 headers for the event published to the DLQ:
- `exc_class`: the name of the error's class, e.g. `RuntimeError`, `ValueError`, etc., which could be used by the DLQ service's processing logic to triage events.
- `exc_info`: the exception text, useful for developers to diagnose problems. Further traceback can be found in the logs using Grafana (if available)

### DIFF
--- a/src/hexkit/providers/akafka/provider/eventsub.py
+++ b/src/hexkit/providers/akafka/provider/eventsub.py
@@ -45,7 +45,7 @@ from hexkit.providers.akafka.provider.utils import (
 
 ORIGINAL_TOPIC_FIELD = "original_topic"
 EXC_CLASS_FIELD = "exc_class"
-EXC_INFO_FIELD = "exc_info"
+EXC_MSG_FIELD = "exc_msg"
 
 
 class ConsumerEvent(Protocol):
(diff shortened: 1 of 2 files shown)


src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -44,6 +44,8 @@
 )
 
 ORIGINAL_TOPIC_FIELD = "original_topic"
+EXC_CLASS_FIELD = "exc_class"
+EXC_INFO_FIELD = "exc_info"
```
Cito, line 48: In Python, [`exc_info`](https://www.oreilly.com/library/view/python-in-a/0596001886/re103.html) has a specific meaning as the 3-tuple of class, value and traceback.

Therefore it would be better to name the field differently, something like `EXC_MESSAGE_FIELD` = `"exc_message"` or `EXC_MSG_FIELD` = `"exc_msg"`.

TheByronHimes: Thanks! I've changed it to "exc_msg". good suggestion

Cito (approved):

mephenor (approved):

PR #137, opened 2024-11-12
Functionality for object md5 in part uploads and multipart copy cleanup

mephenor:
This PR includes 2 parts functionality needed for the upload improvements in DS-Kit an 1 part multipart copy improvements.

- `get_part_upload` now has an optional `part_md5` argument to support the object MD5 functionality
- There is a new `get_object_etag` method, which, for multipart operations, should return the correct MD5 as described here: https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html#large-object-checksums.
Calculating the corresponding checksum currently would happen client side, but could also be moved into hexkit.

- Multipart copies now try to clean up dangling multipart copies. This works as long as we only have one multipart operation at a time for the same object ID, which would currently work based on how the file services are set up. Added an argument to opt out of this behaviour on the client side. Does this need some test cases?

### DIFF
--- a/.pyproject_generation/pyproject_custom.toml
+++ b/.pyproject_generation/pyproject_custom.toml
@@ -53,6 +53,7 @@ legacy_tox_ini = """
 
     [testenv]
     pass_env =
+        CI
         TC_HOST
         DOCKER_HOST
     deps =
--- a/.pyproject_generation/pyproject_custom.toml
+++ b/.pyproject_generation/pyproject_custom.toml
@@ -1,6 +1,6 @@
 [project]
 name = "hexkit"
-version = "4.0.0"
+version = "4.1.0"
 description = "A Toolkit for Building Microservices using the Hexagonal Architecture"
 requires-python = ">=3.9"
 classifiers = [
--- a/.pyproject_generation/pyproject_custom.toml
+++ b/.pyproject_generation/pyproject_custom.toml
@@ -53,7 +53,6 @@ legacy_tox_ini = """
 
     [testenv]
     pass_env =
-        CI
         TC_HOST
         DOCKER_HOST
     deps =
--- a/.pyproject_generation/pyproject_custom.toml
+++ b/.pyproject_generation/pyproject_custom.toml
@@ -1,6 +1,6 @@
 [project]
 name = "hexkit"
-version = "4.1.0"
+version = "4.0.0"
 description = "A Toolkit for Building Microservices using the Hexagonal Architecture"
 requires-python = ">=3.9"
 classifiers = [
--- a/src/hexkit/protocols/objstorage.py
+++ b/src/hexkit/protocols/objstorage.py
@@ -415,7 +415,7 @@ class ObjectStorageProtocol(ABC):
 
     @abstractmethod
     async def _get_object_etag(self, *, bucket_id: str, object_id: str) -> str:
-        """Returns the etag of an object.
+        """Return the etag of an object.
 
         *To be implemented by the provider. Input validation is done outside of this
         method.*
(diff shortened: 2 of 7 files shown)


.pyproject_generation/pyproject_custom.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "hexkit"
-version = "4.0.0"
+version = "4.1.0"
```
Cito, line 3: You do not need to bump the version since 4.0.0 has not yet been released.

src/hexkit/protocols/objstorage.py
```diff
@@ -220,6 +228,7 @@ async def copy_object(
         source_object_id: str,
         dest_bucket_id: str,
         dest_object_id: str,
+        abort_failed: bool = True,
```
Cito, line 231: Can you mention this argument in the docstring?

src/hexkit/protocols/objstorage.py
```diff
@@ -402,6 +413,14 @@ async def _get_object_download_url(
         """
         ...
 
+    @abstractmethod
+    async def _get_object_etag(self, *, bucket_id: str, object_id: str) -> str:
+        """Returns the etag of an object.
```
Cito, line 418:
```suggestion
        """Return the etag of an object.
```

Top line should use imperative mood - https://peps.python.org/pep-0257/#one-line-docstrings

src/hexkit/providers/s3/provider.py
```diff
@@ -757,6 +762,22 @@ async def _get_object_download_url(
 
         return presigned_url
 
+    async def _get_object_etag(self, *, bucket_id: str, object_id: str) -> str:
+        """Returns the etag of an object."""
```
Cito, line 766:
```suggestion
        """Return the etag of an object."""
```

tests/unit/test_federated_s3.py
```diff
@@ -48,6 +48,11 @@ async def test_get_configs_by_alias(federated_s3: FederatedS3Fixture):
 
 async def test_populate_dummy_items(federated_s3: FederatedS3Fixture):
     """Test the populate_dummy_items function on the FederatedS3Fixture."""
+    # Explicitly calling this again, as it seems to solve the Github Actions issue
+    # Needs more investigation on what might be happening during fixture setup that makes this necessary
```
Cito, line 52: I noticed this as well, happens very rarely. Maybe the bucket deletion takes some time?

tests/unit/test_federated_s3.py
```diff
@@ -57,21 +62,18 @@ async def test_populate_dummy_items(federated_s3: FederatedS3Fixture):
     # Populate the items
     await federated_s3.populate_dummy_items(PRIMARY_STORAGE_ALIAS, buckets)
 
+    storage_1 = federated_s3.storages[PRIMARY_STORAGE_ALIAS].storage
+    storage_2 = federated_s3.storages[SECONDARY_STORAGE_ALIAS].storage
```
Cito, line 66: Yes, that's nicer.

Cito (changes requested):
Looks good, some comments belows.

Not sure about the len==1 restriction. Not sure about how difficult it is to test the cleanup - leave it up to you to decide if it's worthwile to add a test.

.pyproject_generation/pyproject_custom.toml
mephenor: Addressed in 0fa7eb3935636c9349c1609fbba698d1558b5417

src/hexkit/protocols/objstorage.py
mephenor: Added some explanation, including when this should be set to `False` in 0fa7eb3935636c9349c1609fbba698d1558b5417

Cito (dismissed):

Cito (approved):

PR #138, opened 2024-11-12
Improve check for bucket existence (GSI-1142)

Cito:
Some S3 operations check the existence of the corresponding bucket as a first step. However, this check currently only works when the client has the permission to list the bucket names.

To avoid the need for this permission, this PR changes the way the existence of the bucket is checked by using a HeadBucket request for the given bucket name instead of getting the list of all existing buckets names.

After this change, the test suite (which contains several tests that rely on this feature) still runs without changes.

Note: This fix should be also backported to V3 after merging.


tests/integration/test_s3.py
```diff
@@ -217,7 +217,7 @@ async def test_handling_non_existing_file_and_bucket(
         )
 
     with pytest.raises(ObjectStorageProtocol.BucketNotFoundError):
-        # copy when source does not exist:
+        # copy when destination does not exist:
```
lkuchenb, line 220: Maybe this could be rephrased to "destination bucket"? Destination objects don't exist in most copy use cases...

lkuchenb (approved):

tests/integration/test_s3.py
Cito: Changed in 4df451e and refined the non-existing source bucket/object case.

mephenor (approved):

PR #141, opened 2024-11-19
Allow colon characters in bucket names

lkuchenb:
Unfortunately there is no straight forward way to test this, since localstack does not permit colon names in bucket names either.

https://github.com/localstack/localstack/blob/e8680021583a49697320ae1d65b5683082d0f00c/localstack-core/localstack/services/s3/utils.py#L499-L503

### DIFF
--- a/.pyproject_generation/pyproject_custom.toml
+++ b/.pyproject_generation/pyproject_custom.toml
@@ -1,6 +1,6 @@
 [project]
 name = "hexkit"
-version = "4.0.0"
+version = "4.1.0"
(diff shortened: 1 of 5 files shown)


TheByronHimes (commented):
We haven't made a 4.0 release yet, so you don't need to increase the version number

Cito (approved):
Looks good.

I pushed a commit that changes the regex only for the S3 provider, makes it a bit more restrictive and adds a simple test.

lkuchenb:
Thank you!

PR #143, opened 2024-11-21
Dynamically define dlq and retry topics (GSI-1175)

TheByronHimes:
### `akafka/config.py`
- Removed the config values for the two topics (`kafka_dlq_topic` and `kafka_retry_topic`). The former is generated by appending `KafkaConfig.service_name` and `-dlq` to the end of the vanilla topic name, with a `.` between. Example: regular topic is 'some-topic', the service name is 'my-service', and the dlq topic that 'my-service' uses for erroring events from 'some-topic' is 'some-topic.my-service-dlq'.
- Removed the validator because the fields it validated are gone

### `akafka/eventsub.py`
- Add a small function for getting the service name from a DLQ topic name
- Dynamically form and subscribe to retry topic based on configured `service_name`.
- Remove `original_topic` header for the first half of the DLQ flow (it's not needed or included when publishing a failed event to the DLQ because the topic name is preserved in the DLQ topic's name)
- In `KafkaEventSubscriber._extract_info()`, it will get the original topic value from the DLQ topic name instead of the old header
- In `KafkaDLQSubscriber`, The retry topic is no longer an init variable because that information is, again, pulled from the service name (which is embedded in the DLQ topic name).

### Tests
- Since the config is simplified, we don't need a bunch of tests to make sure the config validation works.
- Renamed the subscriber instances to either `event_subscriber` or `dlq_subscriber` for clarity
- Moved "test-topic" and "test-type" into constants
- Added a few comments for clarity
- Updated the log assertions that included topic names (previously the test names were `dlq` or `retry`, but these are now `test-topic.test-dlq` or `test-retry`)
- Replaced one use of a translator with an AsyncMock because the translator wasn't actually used (around L590)
- Use constant DLQ and retry topics since the tests all use the same topic and event types

### DIFF
--- a/src/hexkit/providers/akafka/provider/eventsub.py
+++ b/src/hexkit/providers/akafka/provider/eventsub.py
@@ -89,7 +89,7 @@ class ExtractedEventInfo:
 
 def service_name_from_dlq_topic(dlq_topic: str) -> str:
     """Extract the service name from a DLQ topic name."""
-    return dlq_topic.rsplit(".")[1].removesuffix("-dlq")
+    return dlq_topic.rsplit(".", 1)[1].removesuffix("-dlq")
 
 
 def get_event_label(event: ConsumerEvent) -> str:
@@ -683,7 +683,7 @@ class KafkaDLQSubscriber(InboundProviderBase):
                 type_=event.type_,
                 key=event.key,
                 topic=self._retry_topic,
-                headers={ORIGINAL_TOPIC_FIELD: self._dlq_topic.rsplit(".")[0]},
+                headers={ORIGINAL_TOPIC_FIELD: self._dlq_topic.rsplit(".", 1)[0]},
             )
             logging.info(
                 "Published an event with type '%s' to the retry topic '%s'",
--- a/src/hexkit/providers/akafka/provider/eventsub.py
+++ b/src/hexkit/providers/akafka/provider/eventsub.py
@@ -222,13 +222,11 @@ class KafkaEventSubscriber(InboundProviderBase):
         topics = translator.topics_of_interest
 
         if config.kafka_enable_dlq:
-            dlq_topics = [f"{topic}-{config.service_name}-dlq" for topic in topics]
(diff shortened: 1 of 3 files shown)


src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -645,21 +664,26 @@ def __init__(
...
-        original_topic = event.headers[ORIGINAL_TOPIC_FIELD]
 
         async with set_correlation_id(correlation_id):
             await self._publisher.publish(
                 payload=event.payload,
                 type_=event.type_,
                 key=event.key,
                 topic=self._retry_topic,
-                headers={ORIGINAL_TOPIC_FIELD: original_topic},
+                headers={ORIGINAL_TOPIC_FIELD: self._dlq_topic.rsplit(".")[0]},
```
Cito, line 686:
```suggestion
                headers={ORIGINAL_TOPIC_FIELD: self._dlq_topic.rsplit(".", 1)[0]},
```

tests/unit/test_dlqsub.py
```diff
@@ -182,68 +188,36 @@ async def _publish_validated(
...
-    3. If the DLQ is enabled but the topics are not set (either or both)
     """
-    with pytest.raises(ValueError) if error else nullcontext():
-        make_config(
-            retry_topic=retry_topic,
-            dlq_topic=dlq_topic,
-            max_retries=max_retries,
-            enable_dlq=enable_dlq,
-        )
+    max_retries = [-1, 0, 1]
```
Cito, line 212: Maybe parametrize the test instead.

Cito (approved):

src/hexkit/providers/akafka/provider/eventsub.py
TheByronHimes: Thanks. There's also the method at the top that had the same mistake. Fixed

Cito (approved):

PR #144, opened 2024-11-26
Preview DLQ Events (GSI-1127)

TheByronHimes:
### `ExtractedEventInfo`
- The `type_` assignment used to fall back to the header value, but didn't remove it. The result is that the "type" would be found both in the top-level `type_` variable as well as in `headers["type"]`. Now, the value is popped.

### `KafkaDLQSubscriber`
- Removed the subclassing of `InboundProviderBase` because this is actually a specialized class that has little overlap with the normal subscriber provider class.
- `._get_events_from_dlq()`:
  - New method to get the events from the DLQ without hanging. Called by `.preview()`, `.ignore()`, and `.process()`.
  - `max_records` (int) parameter tells the function how many records to retrieve at most.
  - Uses `AIOKafkaConsumer.getmany` with `max_records` and an arbitrary but reasonable timeout (500ms). The purpose of the timeout is to retrieve the records without hanging forever, since we can't know if or how many records will be available to fetch. 
  - This method _does not_ commit offsets itself, because that behavior is dependent on the use case.
- `.preview()`:
  - Returns the next events in the DLQ topic without affecting offsets
  - The value used for `max_records` is `limit + skip`
  - Has 2 parameters: `limit` (int) and `skip` (int). These parameters are for pagination of the preview results.  
`limit` will cap the number of results returned, and `skip` tells the function how far forward to start the preview. If there are no events in the topic, or if `skip` is greater than the number of events available, an empty list is returned.
- `.ignore()`:
  - Replaces `run(ignore=True)`
  - If no events are found in the DLQ, nothing happens.
  - If an event is found, the offsets are committed and nothing more happens. This effectively "ignores" the event.
- `process()`:
  - Replaces `run(ignore=False)`
  - If no events are found in the DLQ, nothing happens.

### Tests
- `TEST_DLQ_EVENT` was created because several tests were defining the same test event to publish directly to the DLQ. Adding this allowed for the condensing of the relevant test cases.
- 3 test functions were added to test the preview functionality
- 1 test function was added to verify the timeout behavior of the methods when there is nothing in the DLQ topic.


tests/unit/test_dlqsub.py
Cito, line 1: Shouldn't this file be under tests/integration rather than tests/unit?

tests/unit/test_dlqsub.py
```diff
@@ -82,7 +94,7 @@ class OutboxDto(BaseModel):
     key="123456",
 )
 
-OUTBOX_EVENT_DELETE = deepcopy(OUTBOX_EVENT_UPSERT)
+OUTBOX_EVENT_DELETE = copy(OUTBOX_EVENT_UPSERT)
```
Cito, line 97: Consider using the (replace](https://docs.python.org/3/library/dataclasses.html#dataclasses.replace) function instead of copying and changing an attribute.

tests/unit/test_dlqsub.py
```diff
@@ -680,14 +674,132 @@ async def process(self, event: ConsumerEvent) -> Optional[ExtractedEventInfo]:
...
+    config = make_config(kafka.config)
+
+    async with set_correlation_id(TEST_CORRELATION_ID):
+        # publish 3 events -- there will be logs about how we supplied correlation_id
+        # but that doesn't matter and can be ignored
+        await kafka.publisher.publish(**vars(TEST_DLQ_EVENT))
+        event2 = copy(TEST_DLQ_EVENT)
+        event3 = copy(TEST_DLQ_EVENT)
+        event2.type_ = "test_type2"
+        event3.type_ = "test_type3"
```
Cito, line 700: see above

Cito (dismissed):
Looks good. See some minor comments.

tests/unit/test_dlqsub.py
TheByronHimes: I wasn't aware of that function, glad you mentioned it. Can think of many times this would have made my life easier!

tests/unit/test_dlqsub.py
TheByronHimes: Probably

src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -705,28 +705,73 @@ async def _handle_dlq_event(self, *, event: ConsumerEvent) -> None:
...
-        """Ignore the event, log it, and commit offsets"""
-        event_label = get_event_label(event)
-        logging.info(
-            "Ignoring event from DLQ topic '%s': %s",
-            self._dlq_topic,
-            event_label,
+    async def _get_events_from_dlq(self, *, max_records: int) -> list[ConsumerEvent]:
+        """Get the next event from the DLQ topic."""
+        fetched = await self._consumer.getmany(  # type: ignore
+            max_records=max_records, timeout_ms=500
```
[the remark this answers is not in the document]
TheByronHimes: Let me know if [79bcb33](https://github.com/ghga-de/hexkit/pull/144/commits/79bcb3395169724d07bf50169cbfead2f43d1008) is what you were thinking.

Cito (approved):

PR #150, opened 2025-01-14
Update how DLQ subscriber is instantiated (GSI-1274)

TheByronHimes:
The use case for the DLQ subscribers involves using many of them at once, for which the construction model of the regular `KafkaEventSubscriber` is ill-suited. This PR ditches the `construct` method in favor of a normal `__init__` and explicit `start()`, `stop()`. I also included the async context manager functions (`__aenter__()`/`__aexit__()`) to call start and stop since this is helpful for tests where there might only be one subscriber.

I also updated `test_dlqsub.py` to use the blanket `pytestmark` since there was only one sync function.

### DIFF
--- a/src/hexkit/providers/akafka/provider/eventsub.py
+++ b/src/hexkit/providers/akafka/provider/eventsub.py
@@ -566,10 +566,8 @@ class KafkaDLQSubscriber:
     DLQEventProcessor definition.
     """
 
-    @classmethod
-    @asynccontextmanager
-    async def construct(  # noqa: PLR0913
-        cls,
+    def __init__(  # noqa: PLR0913
+        self,
         *,
         config: KafkaConfig,
         dlq_topic: str,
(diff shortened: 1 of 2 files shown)


src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -722,9 +695,9 @@ async def _handle_dlq_event_manual(
...
         if dlq_event.headers["correlation_id"] != override.headers["correlation_id"]:
             msg = (
-                "Cannot manually resolve DLQ event due to correlation ID mismatch.\n"
-                + f"User-supplied event ID: {override.headers['correlation_id']}\n"
-                + f"DLQ event ID: {dlq_event.headers['correlation_id']}"
+                "Cannot manually resolve DLQ event due to correlation ID mismatch."
+                + f"\nExpected {dlq_event.headers['correlation_id']} but user"
+                + f" gave {override.headers['correlation_id']}"
             )
             raise RuntimeError(msg)
```
mephenor, line 702: Is this expected to be handled somewhere or is this unrecoverable?

tests/integration/test_dlqsub.py
```diff
@@ -856,11 +842,15 @@ async def test_process_override_different_cid(kafka: KafkaFixture):
     )
 
     # Create the DLQ subscriber and manually resolve the DLQ event, expecting an error
-    msg = "Cannot manually resolve DLQ event because the correlation ID doesn't match."
-    async with KafkaDLQSubscriber.construct(
+    msg = (
+        "Cannot manually resolve DLQ event due to correlation ID mismatch."
+        + f"\nExpected {TEST_CORRELATION_ID} but user gave {headers['correlation_id']}"
+    )
```
mephenor, line 848: See first comment

tests/integration/test_dlqsub.py
```diff
@@ -856,11 +842,15 @@ async def test_process_override_different_cid(kafka: KafkaFixture):
...
+    async with KafkaDLQSubscriber(
         config=config,
         dlq_topic=TEST_DLQ_TOPIC,
         dlq_publisher=DummyPublisher(),
     ) as dlq_subscriber:
-        with pytest.raises(ValueError, match=msg):
+        with pytest.raises(DLQProcessingError) as err:
```
mephenor, line 854: Where is this error type defined and does it wrap the generic RuntimeError?

src/hexkit/providers/akafka/provider/eventsub.py
TheByronHimes: Unrecoverable - it's captured and re-raised as a DLQProcessingError, but it's meant as a safeguard. All DLQ events should have a correlation ID, and we shouldn't be modifying that. So if we supply the fixed event directly, the CID should match.

tests/integration/test_dlqsub.py
TheByronHimes: It does wrap `RuntimeError`, and it's defined in the same module as the Kafka subscribers and the code is located between the two subscriber classes ([here](https://github.com/ghga-de/hexkit/blob/fcc3c0dd784680df80fb08ace8a9e4e911664ce1/src/hexkit/providers/akafka/provider/eventsub.py#L518), specifically).

mephenor (approved):

PR #163, opened 2025-03-26
Aiokafka extended OpenTelemetry support

mephenor:
This PR adds some OpenTelemetry specific code to properly support cross service context propagation using AIOKafka.
This is needed due to the combination of two specific circumstances:
1) The aiokafka instrumentation library returns to the previously opened context and does not open a new child span. Thus a parent span using the propagated context is needed to link between the incoming event context and downstream spans.
2) The parent span needs to be opened before event consumption, but needs access to the headers to extract the context. Service code does not have access to Kafka headers, so this has to happen in hexkit.

Lines 650+ in `event_sub` are unchanged, just indented due to to context manager use.

### DIFF
--- a/src/hexkit/providers/akafka/provider/eventsub.py
+++ b/src/hexkit/providers/akafka/provider/eventsub.py
@@ -645,7 +645,7 @@ class KafkaEventSubscriber(InboundProviderBase):
             getter=otel_extractor,
         )
         with tracer.start_as_current_span(
-            name="Consume event", context=extracted_context
+            name="KafkaEventSubscriber._consume_event", context=extracted_context
         ):
             event_id = get_event_id(event, service_name=self._service_name)
             event_info = self._extract_info(event)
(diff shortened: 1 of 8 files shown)


src/hexkit/providers/akafka/provider/eventsub.py
```diff
@@ -610,39 +639,49 @@ def _validate_extracted_info(self, event: ExtractedEventInfo):
...
-            # Always acknowledge event receipt for ignored events
-            await self._consumer.commit()
-            return
+        otel_extractor = AIOKafkaOtelContextExtractor()
+        extracted_context = extract(
+            event.headers,
+            getter=otel_extractor,
+        )
+        with tracer.start_as_current_span(
+            name="Consume event", context=extracted_context
```
Cito, line 648: Maybe it's better to not use spaces in span names, i.e. "consume_event" instead?

https://oneuptime.com/blog/post/2024-11-04-how-to-name-spans-in-opentelemetry/view

Cito (approved):
Looks good, just one comment.

src/hexkit/providers/akafka/provider/eventsub.py
mephenor: Thanks for the link 
It seems I also need to change my span naming strategy in the file services

PR #167, opened 2025-04-14
Expose singleton kafka client and clear compacted topics

TheByronHimes:
This PR lets the kafka fixture work with compacted topics when running `clear_topics()`. 
The default `cleanup.policy` is `delete`, so topics are only ever compacted by manually setting them that way. To that end, there is a `.set_topic_config()` (for general config updates) and a focused `.set_cleanup_policy()` method to conveniently set that specific piece of config for a topic. 

The reason for this PR is that testing in the SMS (specifically for the DLQ) can't proceed without this functionality. It could be hacked together in the SMS or wherever, but the Kafka test fixture also calls `clear_topics()` for the setup portion of most tests. If there are compacted topics lying around after a previous test, then it will (until now) get an error.

I changed the first test in `test_akafka_testutils.py` because it was calling `clear_topics` after publishing the target message but while the consumer was still running. So the message would often end up in the hands of the consumer before the topic was cleared.


src/hexkit/providers/akafka/testutils.py
```diff
@@ -453,15 +528,32 @@ async def clear_topics(
...
+
+            # Record the current cleanup policies before modifying them
+            for topic in topics:
+                policy = await self.get_cleanup_policy(topic=topic)
+                if policy and "compact" in policy.split(","):
+                    config = await self.get_all_topic_config(topic=topic)
+                    original_configs[topic] = config
+                    await self.set_topic_config(
+                        topic=topic, config={"retention.ms": -1}
+                    )
```
mephenor, line 554: Some comment explaining why this is done would be great.
Is this always set in the original config, i.e. will this be reset to the correct value later on?

TheByronHimes: This was recommended in an article I found, but I think its actually not useful here because the default retention value is much much longer than any test we'll run. But hypothetically, it would prevent any time-based deletions from occurring and interfering with the process. I don't understand it more than that. However, I took it out because it wasn't needed in the end.

src/hexkit/providers/akafka/testutils.py
```diff
@@ -442,6 +445,78 @@ def record_events(
...
+        return await self.get_topic_config(topic=topic, config_name="cleanup.policy")
+
+    async def get_topic_config(self, *, topic: str, config_name: str) -> Optional[str]:
+        """Get the current config for a topic or None if the topic/config doesn't exist"""
+        async with self.get_admin_client() as admin_client:
+            config_response = await admin_client.describe_configs(
+                [ConfigResource(ConfigResourceType.TOPIC, name=topic)]
+            )
+        config_resources = config_response[0].resources[0]
+        configs = config_resources[-1]
```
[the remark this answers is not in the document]
TheByronHimes: I have done so in [db2bac2](https://github.com/ghga-de/hexkit/pull/167/commits/db2bac28f704ef6487a620e2f558e4c4d30f0452)

mephenor (approved):

PR #169, opened 2025-04-15
OpenTelemetry Tracer decorator + set default exporter protocol to http/protobuf 

mephenor:
Adds a custom `SpanTracer` class that holds the tracer instance and provides a decorator method that automatically starts a span and populates its name with the function `__qualname__`.

Also extends `configure_tracer` to allow to set a default exporter protocol programmatically like the `opentelemetry-distro` library does, but instead of setting `grpc` as default, `http/protobuf` is set.

### DIFF
--- a/src/hexkit/opentelemetry_setup.py
+++ b/src/hexkit/opentelemetry_setup.py
@@ -44,20 +44,21 @@ class SpanTracer:
         self.tracer = trace.get_tracer(name)
 
     def start_span(
-        self,
-        function: Callable,
-        record_exception: bool = False,
-        set_status_on_exception: bool = False,
+        self, *, record_exception: bool = False, set_status_on_exception: bool = False
     ):
         """Decorator function starting a span populated with the __qualname__ of the wrapped function"""
 
-        def traced_function(*args, **kwargs):
-            name = function.__qualname__
-            with self.tracer.start_as_current_span(
-                name,
-                record_exception=record_exception,
-                set_status_on_exception=set_status_on_exception,
-            ):
-                return function(*args, **kwargs)
+        def outer_wrapper(function: Callable):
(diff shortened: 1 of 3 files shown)


src/hexkit/opentelemetry_setup.py
```diff
@@ -14,17 +14,41 @@
...
 from opentelemetry.sdk.resources import SERVICE_NAME, Resource
 from opentelemetry.sdk.trace import TracerProvider
 from opentelemetry.sdk.trace.export import BatchSpanProcessor
 
 
-def configure_tracer(service_name: str):
-    """Set up a global tracer for a specific service."""
+def configure_tracer(service_name: str, protocol: str = "http/protobuf"):
+    """Set up a global tracer for a specific service using the given exporter protocol."""
+    # opentelemetry distro sets this to grpc, but in the current context http/protobuf is prefered
```
mephenor, line 30:
```suggestion
    # opentelemetry distro sets this to grpc, but in the current context http/protobuf is preferred
```

src/hexkit/opentelemetry_setup.py
```diff
@@ -14,17 +14,41 @@
...
+
+
+class SpanTracer:
+    """Custom tracer class providing a decorator to autpopulate span names."""
+
+    def __init__(self, name):
+        self.tracer = trace.get_tracer(name)
+
+    def start_span(self, function: Callable):
+        """Decorator function starting a span populated with the function __qualname__"""
```
mephenor, line 47:
```suggestion
        """Decorator function starting a span populated with the __qualname__ of the wrapped function"""
```

TheByronHimes (dismissed):

TheByronHimes (approved):

PR #171, opened 2025-05-05
OpenTelemetry Refactoring

mephenor:
This PR adds a `start_span` decorator wrapping `SpanTracer.start_span`, allowing to populate the Tracer in one central place and still use the decorator as intended.
A new OpenTelemtry specific config class is provided. 

Additionally, the config code now disables both the metrics and log exporter as the current service implementations only care about traces.
A flag can also be passed to disable OpenTelemetry instead, which should also disable autoinstrumentation, but this needs some testing, as additional env vars might be needed to fully realize this.

The static sampler has been replaced by a probabilistic sampler respecting parent sampling decisions, so only full traces within a service will be either sampled or not.
With the default value of 1, there should be no difference in behaviour to the previous default implementation, but by exposing this, services can be quickly configured to only export a subset of traces, reducing load if OpenTelemetry causes to much of an overhead.
This is a simple head sampling approach and the sampled traces might not be representative.


src/hexkit/opentelemetry_setup.py
```diff
@@ -61,3 +66,105 @@ def traced_function(*args, **kwargs):
...
+
+
+class OpenTelemetryConfig(BaseSettings):
+    """OpenTelemetry specific configuration options"""
+
+    enable_opentelemetry: bool = Field(
+        default=False,
+        description="If set to true, this will run necessary setup code."
+        "If set to false, environment variables are set that should also effectively "
+        "disable autoinstrumentation",
```
mephenor, line 78:
```suggestion
        "disable autoinstrumentation.",
```

src/hexkit/opentelemetry_setup.py
```diff
@@ -61,3 +66,105 @@ def traced_function(*args, **kwargs):
...
+
+    Setup of the TracerProvider is done programmatically, all other configuration exports
+    OpenTelemetry specific environment variables.
+    """
+    global TRACER
+    # opentelemetry distro sets this to grpc, but in the current context http/protobuf is preferred
+    os.environ.setdefault(OTEL_EXPORTER_OTLP_PROTOCOL, config.protocol)
+    # Disable OpenTelemetry metrics and logs explicitly as they are not processed in the backend currently
+    # This overwrites the defaults of `otlp` set in opentelemetry distro
+    os.environ.setdefault(OTEL_METRICS_EXPORTER, "none")
```
Cito, line 116: Actually, if the distro sets `OTEL_METRICS_EXPORTER="otlp"` then the `.setdefault` will not change it. Is this what we want? Or do we rather want to unconditionally set `os.environ[OTEL_METRICS_EXPORTER] = "none"`?

src/hexkit/opentelemetry_setup.py
```diff
@@ -61,3 +66,105 @@ def traced_function(*args, **kwargs):
...
+        trace_provider = TracerProvider(resource=resource, sampler=sampler)
+        processor = BatchSpanProcessor(OTLPSpanExporter())
+        trace_provider.add_span_processor(processor)
+        trace.set_tracer_provider(trace_provider)
+        TRACER = SpanTracer(service_name)
+    else:
+        # Currently OTEL_SDK_DISABLED doesn't seem to be honored by all implementations yet
+        # It seems to be working well enough for the Python implementation, but to be on
+        # the safe side, let's explicitly disable the trace exporter for now
+        os.environ.setdefault(OTEL_TRACES_EXPORTER, "none")
```
Cito, line 146: see above

Cito (approved):
Looks good, just one remark/question.

src/hexkit/opentelemetry_setup.py
mephenor: Thanks for catching that, copied that from a testing context and didn't think about it not overwriting what's already set. I'll adjust that

PR #179, opened 2025-06-02
Add Kafka Producer Compression (GSI-1669)

TheByronHimes:
This PR adds opt-in Kafka producer compression. We can only set the producer compression through `hexkit`, and the matching external change will be to set compression.type=producer in the broker config for Strimzi.


src/hexkit/providers/akafka/config.py
```diff
@@ -117,3 +119,15 @@ class KafkaConfig(BaseSettings):
         title="Kafka Retry Backoff",
         examples=[0, 1, 2, 3, 5],
     )
+    kafka_compression_type: Optional[KafkaCompressionType] = Field(
```
Cito, line 122: Can you move this setting next to `kafka_max_message_size`, since they are a bit connected?

And in the `kafka_max_message_size` description, mention that this refers to the uncompressed size
and enlarging this can be combined with adding compression.

src/hexkit/providers/akafka/config.py
```diff
@@ -117,3 +119,15 @@ class KafkaConfig(BaseSettings):
...
+        default=None,
+        description=(
+            "The compression type used for messages. Valid values are: None, gzip,"
+            + " snappy, lz4, and zstd. If None, no compression is applied. This"
+            + " setting is only relevant for the producer and has no effect on the"
+            + " consumer. If set to a value, the producer will compress messages before"
+            + " sending them to the Kafka broker."
```
Cito, line 129: Recommend zstd for topics with large messages?

PR #180, opened 2025-06-11
Fix OTel autoinstrumentation not working

mephenor:
This PR addresses autoinstrumentation spans missing after the last release.
A custom no-op `HexkitDistro` class has been implemented that provides entrypoints for autoinstrumentation via the default implementation of `DefaultDistro`.
The distro is exposed in `pyproject.toml` via `[project.entry-points.opentelemetry_distro]`.
This will need some manual testing, especially for the case where `configure_opentelemetry` is not called explicitly, as it might enable autoinstrumentation unintentionally in those cases and there might be consequences e.g. crashes as no defaults are set for exporter endpoints, etc.

Setting `os.environ[OTEL_SDK_DISABLED] = "true"` in the `_configure` method of the distro might be a valid solution to address this, though that depends on if everything works correctly when trying to set it to `false` later on in `configure_opentelemetry`.

The default for `record_exception` and  `set_status_on_exception` of the span decorator has been reverted to the OpenTelemetry default of `True`.


src/hexkit/opentelemetry.py
```diff
@@ -95,6 +98,17 @@ class OpenTelemetryConfig(BaseSettings):
...
+class HexkitDistro(BaseDistro):
+    """
+    Custom OpenTelemetry Distro configuring a default set of configuration
+    based on a config file.
+    """
+
+    def _configure(self, **kwargs):
+        """Custom DefaultDistro iplementation to provide autoinstrumentation entry points."""
+        # disable by default, opt into via `configure_opentelemetry`
+
```
mephenor, line 110:
```suggestion

```

TheByronHimes (approved):

PR #188, opened 2025-07-10
Replace motor with pymongo async (GSI-1727)

TheByronHimes:
Motor is on the way out (though we have some time). I chose to go ahead and move to pymongo async, though. 

This PR removes motor as a dependency and replaces it with pymongo.

### Changes
- Adds a new sync/async context manager, `ConfiguredMongoClient`, that can be used to automatically configure and open/close mongo clients. 
  - Using `async with` yields an `AsyncMongoClient`, while just `with...` yields a synchronous `MongoClient`. Both are cleaned up on exit.
  - Pymongo Async produces loads of warnings in tests if you don't clean up the client connections, so this CM will keep things tidy.
- Updates `MongoDbDaoFactory` with a `.construct()` method, and adds a required `client` parameter to the `__init__()` method. This means we'll have to switch all the instances of direct instantiation (`dao_factory = MongoDbDaoFactory(config=config)`) to instead be:
```python
async with MongoDbDaoFactory.construct(config=config) as dao_factory:
    pass
    
# alternatively:
async with ConfiguredMongoClient(config=config) as client:
    dao_factory = MongoDbDaoFactory(config=config, client=client)
```
  - As far as I could tell, all our usage of MongoDbDaoFactory occurs in async methods, so this should work fine. It will clean up the client when done.

### DIFF
--- a/src/hexkit/providers/mongodb/testutils.py
+++ b/src/hexkit/providers/mongodb/testutils.py
@@ -31,9 +31,9 @@ from testcontainers.mongodb import MongoDbContainer
 
 from hexkit.custom_types import PytestScope
 from hexkit.providers.mongodb.provider import (
+    ConfiguredMongoClient,
     MongoDbConfig,
     MongoDbDaoFactory,
-    get_configured_mongo_client,
 )
 
 MONGODB_IMAGE = "mongo:7.0.15"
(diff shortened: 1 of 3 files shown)


src/hexkit/providers/mongodb/testutils.py
```diff
@@ -125,25 +125,24 @@ def get_mongodb_container_fixture(
...
-
-    # Instead of `.get_connection_client()`, manually create it with correct codec opts
-    client = get_configured_mongo_client(config=config, client_cls=MongoClient)
-    yield MongoDbFixture(
-        client=client,
-        config=config,
-        dao_factory=dao_factory,
-    )
+    async with MongoDbDaoFactory.construct(config=config) as dao_factory:
+        # Instead of `.get_connection_client()`, manually create it with correct codec opts
```
Cito, line 137: which `get_connection_client` method does this refer to?

Cito (approved):
LGTM, just made two comments.

src/hexkit/providers/mongodb/testutils.py
TheByronHimes: The one from the testcontainer

Cito: Ok, that should be mentioned. And a reason why we do this instead. But I saw you removed it already, that's ok.

PR #190, opened 2025-07-14
Helper functions for hexkit v6 migrations (GSI-1730)

TheByronHimes:
As we roll out hexkit v6 to services, there will be a lot of similar migration logic happening. To make the rollout smoother, this PR adds a module in the mongodb migrations subpackage to perform common operations, like converting isoformat string datetimes to actual datetimes. The functions are very clearly named and have been unit tested. We can delete them if we want to in some future hexkit version.

### DIFF
--- a/src/hexkit/providers/mongodb/migrations/helpers.py
+++ b/src/hexkit/providers/mongodb/migrations/helpers.py
@@ -21,6 +21,7 @@ from typing import Optional
 from uuid import UUID, uuid4
 
 from hexkit.providers.mongodb.migrations import Document
+from hexkit.utils import round_datetime_to_ms
 
 __all__ = [
     "convert_outbox_correlation_id_v6",
@@ -53,12 +54,7 @@ def convert_uuids_and_datetimes_v6(
             old_dt = datetime.fromisoformat(doc[field])
             if old_dt.tzname() != "UTC":
                 old_dt = old_dt.astimezone(timezone.utc)
-            microseconds = old_dt.microsecond
-
-            # Round to milliseconds and account for case where microseconds are >999500
-            rounded_ms = (round(microseconds / 1000) % 1000) * 1000
-            new_dt = old_dt.replace(microsecond=rounded_ms)
-            doc[field] = new_dt
+            doc[field] = round_datetime_to_ms(old_dt)
         return doc
 
     return convert
--- a/src/hexkit/providers/mongodb/migrations/helpers.py
+++ b/src/hexkit/providers/mongodb/migrations/helpers.py
@@ -36,6 +36,10 @@ def convert_uuids_and_datetimes_v6(
 ) -> Callable[[Document], Awaitable[Document]]:
     """Produce a function to convert a document to the format expected by Hexkit v6.
 
+    If `uuid_fields` is provided, it will convert those fields from string to UUID.
+    If `date_fields` is provided, it will convert those fields from isoformat strings
+    to UTC datetime objects with the microseconds rounded to milliseconds.
+
     Args:
         uuid_fields: List of fields currently storing UUIDs in string format.
         date_fields: List of fields currently storing datetimes as isoformat strings.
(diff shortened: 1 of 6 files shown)


src/hexkit/providers/mongodb/migrations/helpers.py
```diff
@@ -0,0 +1,91 @@
...
+from hexkit.providers.mongodb.migrations import Document
+
+__all__ = [
+    "convert_outbox_correlation_id_v6",
+    "convert_persistent_event_v6",
+    "convert_uuids_and_datetimes_v6",
+]
+
+
+def convert_uuids_and_datetimes_v6(
```
mephenor, line 32: There's also uuid v6, this could be slightly ambiguous

src/hexkit/providers/mongodb/migrations/helpers.py
```diff
@@ -0,0 +1,91 @@
...
+    """Produce a function to convert a document to the format expected by Hexkit v6.
+
+    Args:
+        uuid_fields: List of fields currently storing UUIDs in string format.
+        date_fields: List of fields currently storing datetimes as isoformat strings.
+    """
+
+    async def convert(doc: Document) -> Document:
+        """Convert the document."""
+        for field in uuid_fields or []:
```
mephenor, line 46: TIL, never knew this is possible.
Do you have a link to an eplanation on how this is evaluated?

src/hexkit/providers/mongodb/migrations/helpers.py
```diff
@@ -0,0 +1,91 @@
...
+            doc[field] = UUID(doc[field])
+        for field in date_fields or []:
+            old_dt = datetime.fromisoformat(doc[field])
+            new_dt = old_dt.replace(microsecond=old_dt.microsecond // 1000 * 1000)
```
mephenor, line 50: Why divide and multiply by the same value here, truncating precision?

src/hexkit/providers/mongodb/migrations/helpers.py
```diff
@@ -0,0 +1,91 @@
...
+        uuid_fields: List of fields currently storing UUIDs in string format.
+        date_fields: List of fields currently storing datetimes as isoformat strings.
+    """
+
+    async def convert(doc: Document) -> Document:
+        """Convert the document."""
+        for field in uuid_fields or []:
+            doc[field] = UUID(doc[field])
+        for field in date_fields or []:
+            old_dt = datetime.fromisoformat(doc[field])
```
mephenor, line 49: Would it make sense to check if the timezone is set after parsing, so raise if we encounter a datetime that might be naive?

src/hexkit/providers/mongodb/migrations/helpers.py
TheByronHimes: MongoDB only has millisecond precision. To make our in-memory data match the database, we will create timestamps with that same level of precision. Python's datetimes don't have a separate field for milliseconds, only one for microseconds. So this operation effectively removes anything below the microseconds. Maybe a more accurate version though would be `=round(old_dt.microsecond / 1000) * 1000`? What do you think?

src/hexkit/providers/mongodb/migrations/helpers.py
TheByronHimes: It's a ternary shortcut, not sure of the real name.
[random site](https://docs.sourcery.ai/Coding-Assistant/Reference/Rules-and-In-Line-Suggestions/Python/Default-Rules/use-or-for-fallback/)
[real python](https://realpython.com/python-or-operator/#non-boolean-contexts)

src/hexkit/providers/mongodb/migrations/helpers.py
TheByronHimes: See the changes made in [faea69c](https://github.com/ghga-de/hexkit/pull/190/commits/faea69c8a901e9c29904a776fcdf239d9d26c6f1). It will now convert to UTC if naive or not UTC.

src/hexkit/providers/mongodb/migrations/helpers.py
TheByronHimes: Please see [991b157](https://github.com/ghga-de/hexkit/pull/190/commits/991b157d0d8827bd18f39274342607e7d9cb6f5c)

src/hexkit/utils.py
```diff
@@ -127,6 +127,20 @@ async def set_context_var(context_var: ContextVar, value: Any):
...
+    microseconds = dt.microsecond
+    sub_ms = microseconds % 1000
+    if not sub_ms:
+        return dt
+
+    # Calculate the delta to add or subtract to round to the nearest millisecond.
+    # we subtract sub_ms either from 1000 if rounding up or from 0 if rounding down
+    delta = timedelta(microseconds=(1000 * (sub_ms >= 500)) - sub_ms)
+    new_dt = dt + delta
+    return new_dt
```
mephenor, line 141: Could also collapse that into one line

PR #198, opened 2025-10-01
Add basic documentation

Cito:


TheByronHimes (approved):
Thanks for putting this together. There are many sections that have merely a heading without any text, but I don't think they all need a disclaimer about being a WIP because that's obvious and we're the only ones reading it at the moment.

PR #227, opened 2026-04-20
Support ListParts method for object storage (GSI-2311)

TheByronHimes:
Adds a new `list_parts()` method to the `ObjectStorageProtocol` and the `S3ObjectStorage` classes.

In addition to bucket/object/upload IDs, the method optionally accepts `max_parts` and `first_part_no`. 

If `max_parts` is specified, the results are capped at that amount.

If `first_part_no` is specified, the list of returned parts will start with that part number, if possible.

`ValueError` is raised if either `max_parts` or `first_part_no` are specified with invalid arguments.

Bumps hexkit to version `8.2.0`

### DIFF
--- a/src/hexkit/providers/s3/provider/objstorage.py
+++ b/src/hexkit/providers/s3/provider/objstorage.py
@@ -370,6 +370,58 @@ class S3ObjectStorage(ObjectStorageProtocol):
 
         return uploads
 
+    async def _list_parts(
+        self,
+        *,
+        bucket_id: str,
+        object_id: str,
+        upload_id: str,
+        max_parts: int | None = None,
+        first_part_no: int | None = None,
+    ) -> list[dict]:
+        """Lists the parts that have been uploaded for a specific multipart upload."""
+        await self._assert_multipart_upload_exists(
+            upload_id=upload_id, bucket_id=bucket_id, object_id=object_id
+        )
+
+        params: dict[str, Any] = {
+            "Bucket": bucket_id,
+            "Key": object_id,
+            "UploadId": upload_id,
+        }
+        if max_parts:
+            params["MaxParts"] = max_parts
+        if first_part_no:
+            # S3 returns parts starting *after* PartNumberMarker, so subtract 1 to get
+            #  more intuitive behavior where the first part returned is first_part_no
+            params["PartNumberMarker"] = first_part_no - 1
+
+        try:
+            if max_parts is not None:
+                # If max_parts is specified, make single call
+                response = await asyncio.to_thread(self._client.list_parts, **params)
+                return response.get("Parts", [])
+
+            # If max_parts isn't specified, paginate manually to collect all parts
+            parts: list[dict] = []
+            while True:
+                response = await asyncio.to_thread(self._client.list_parts, **params)
+                parts.extend(response.get("Parts", []))
+
+                # ListParts is limited to 1000 parts per paginated request, so repeat
+                #  the call until all parts are received (IsTruncated=False)
+                if not response.get("IsTruncated"):
+                    break
+                params["PartNumberMarker"] = response["NextPartNumberMarker"]
+            return parts
+        except botocore.exceptions.ClientError as error:
+            raise self._translate_s3_client_errors(
+                error,
+                upload_id=upload_id,
+                bucket_id=bucket_id,
+                object_id=object_id,
+            ) from error
+
     async def _get_all_multipart_uploads(self, *, bucket_id: str) -> dict[str, str]:
         """Gets all active multipart uploads for the given bucket ID.
 
@@ -506,54 +558,19 @@ class S3ObjectStorage(ObjectStorageProtocol):
                 object_id=object_id,
             ) from error
 
-    async def _get_parts_info(
-        self,
-        *,
-        upload_id: str,
-        bucket_id: str,
-        object_id: str,
-    ) -> dict:
-        """Get information on parts uploaded as part of the specified multi-part upload."""
-        await self._assert_multipart_upload_exists(
-            upload_id=upload_id, bucket_id=bucket_id, object_id=object_id
-        )
-
-        try:
-            response_iter = await asyncio.to_thread(
-                self._client.get_paginator("list_parts").paginate,
-                Bucket=bucket_id,
-                Key=object_id,
-                UploadId=upload_id,
-            )
-            complete_response: dict[str, Any] = {}
-            for response_page in response_iter:
-                if not complete_response:
-                    complete_response = response_page
-                    continue
-                complete_response["Parts"].extend(response_page["Parts"])
-            return complete_response
-        except botocore.exceptions.ClientError as error:
-            raise self._translate_s3_client_errors(
-                error,
-                upload_id=upload_id,
-                bucket_id=bucket_id,
-                object_id=object_id,
-            ) from error
-
     async def _check_uploaded_parts(
         self,
         *,
         upload_id: str,
         bucket_id: str,
         object_id: str,
-        parts_info: dict,
+        parts_info: list[dict],
         anticipated_part_quantity: int | None = None,
         anticipated_part_size: int | None = None,
     ) -> None:
         """Check size and quantity of parts"""
         # check the part quantity:
-        parts = parts_info.get("Parts")
-        if parts is None or len(parts) == 0:
+        if not parts_info:
             raise self.MultiPartUploadConfirmError(
                 upload_id=upload_id,
                 bucket_id=bucket_id,
@@ -561,7 +578,7 @@ class S3ObjectStorage(ObjectStorageProtocol):
                 reason="Zero parts received.",
             )
 
-        part_quantity = len(parts)
+        part_quantity = len(parts_info)
         if (
             anticipated_part_quantity is not None
             and part_quantity != anticipated_part_quantity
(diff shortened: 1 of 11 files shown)


src/hexkit/providers/s3/provider/objstorage.py
```diff
@@ -370,6 +370,78 @@ async def _list_multipart_uploads_for_object(
...
+            "Key": object_id,
+            "UploadId": upload_id,
+        }
+        if max_parts is not None:
+            if isinstance(max_parts, int) and max_parts > 0:
+                params["MaxParts"] = max_parts
+            else:
+                raise ValueError(f"{max_parts} is not a valid argument for max_parts")
+        if first_part_no is not None:
+            if isinstance(first_part_no, int) and first_part_no > 0:
```
mephenor: Why do the instance check here if it should be an int in the non null case via the method signature?
The usual `is not None` is easier to parse.

src/hexkit/providers/s3/provider/objstorage.py
```diff
@@ -370,6 +370,78 @@ async def _list_multipart_uploads_for_object(
...
+                # S3 returns parts starting *after* PartNumberMarker, so subtract 1 to get
+                #  more intuitive behavior where the first part returned is first_part_no
+                params["PartNumberMarker"] = first_part_no - 1
```
mephenor, line 417: Tests should check `first_part_no=1`

src/hexkit/providers/s3/provider/objstorage.py
```diff
@@ -370,6 +370,78 @@ async def _list_multipart_uploads_for_object(
...
+            else:
+                raise ValueError(
+                    f"{first_part_no} is not a valid argument for first_part_no"
+                )
+
+        try:
+            if max_parts is not None:
+                # If max_parts is specified, make single call
+                response = await asyncio.to_thread(self._client.list_parts, **params)
+                return response.get("Parts", [])
```
mephenor: Missing pagination for this option should be reported in the doc string, so there are no surprises when the results only contain the first page.

src/hexkit/providers/s3/provider/objstorage.py
```diff
@@ -370,6 +370,78 @@ async def _list_multipart_uploads_for_object(
...
+            # If max_parts isn't specified, paginate manually to collect all parts
+            parts: list[dict] = []
+            while True:
+                response = await asyncio.to_thread(self._client.list_parts, **params)
+                parts.extend(response.get("Parts", []))
+
+                # ListParts is limited to 1000 parts per paginated request, so repeat
+                #  the call until all parts are received (IsTruncated=False)
+                if not response.get("IsTruncated"):
+                    break
```
mephenor: This will be set to false on the last page because it contains less than (or equal) the max amount of items?

src/hexkit/providers/s3/provider/objstorage.py
```diff
@@ -506,62 +578,27 @@ async def _get_part_upload_url(
...
-            ) from error
-
     async def _check_uploaded_parts(
         self,
         *,
         upload_id: str,
         bucket_id: str,
         object_id: str,
-        parts_info: dict,
+        parts_info: list[dict],
```
mephenor: Might be helpful to document the format in the docstring if looking at this in isolation.
Had to recheck the code above to realize we no longer merge response dicts, but just collect them in a list.

mephenor: Could also rename this to just `parts`, as it's no longer populated from the result of `get_parts_info` and is an internal method either way.

src/hexkit/providers/testing/objstorage.py
```diff
@@ -178,6 +178,44 @@ async def _list_multipart_uploads_for_object(
...
+        and the return value is dummy data only.
+        """
+        await self._assert_bucket_exists(bucket_id)
+        await self._assert_multipart_upload_exists(
+            upload_id=upload_id,
+            bucket_id=bucket_id,
+            object_id=object_id,
+            assert_exclusiveness=False,
+        )
+        if max_parts is not None and not (isinstance(max_parts, int) and max_parts > 0):
```
mephenor: The middle assertion isn't necessary.

src/hexkit/providers/testing/objstorage.py
```diff
@@ -178,6 +178,44 @@ async def _list_multipart_uploads_for_object(
...
+            raise ValueError(f"{max_parts} is not a valid argument for max_parts")
+        if first_part_no is not None and not (
+            isinstance(first_part_no, int) and first_part_no > 0
```
mephenor: See above

tests/integration/providers/objstorage/test_s3_objstorage.py
```diff
@@ -377,6 +377,100 @@ async def test_md5_in_part_url(s3: S3Fixture):
     assert "content-md5=dummy-md5" in url
 
 
+async def test_list_parts(s3: S3Fixture):
```
mephenor, line 380: If we don't paginate on max_parts, this should also check if the correct, truncated page is returned when `first_part_no` is set, i.e. check what happens if you have more elements than will be returned by the unpaginated response.

tests/integration/providers/objstorage/test_s3_objstorage.py
```diff
@@ -377,6 +377,100 @@ async def test_md5_in_part_url(s3: S3Fixture):
...
+    )
+
+    # Check parts list again, should get empty list
+    no_parts = await s3.storage.list_parts(
+        bucket_id=bucket_id, object_id="throwaway", upload_id=uid
+    )
+    assert not no_parts
+
+
+async def test_list_parts_invalid_args(s3: S3Fixture):
```
mephenor, line 444: What happens if `max_parts` or `first_part_no` is set to a value higher than the amount of actual parts?

tests/unit/test_in_mem_objstorage.py
mephenor, line 1: Comments from the other test module also apply here.

mephenor (changes requested):

src/hexkit/providers/s3/provider/objstorage.py
TheByronHimes: That's what I understand from [here](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListParts.html#API_ListParts_ResponseElements)

tests/integration/providers/objstorage/test_s3_objstorage.py
TheByronHimes: In the case of max_parts, nothing special. The results don't exceed the limit and that's it.
In the case of first_part_no, the list is empty.

tests/integration/providers/objstorage/test_s3_objstorage.py
```diff
@@ -470,6 +470,56 @@ async def test_list_parts_invalid_args(s3: S3Fixture):
...
+        parts = await s3.storage.list_parts(
+            bucket_id=bucket_id,
+            object_id=object_id,
+            upload_id=upload_id,
+            max_parts=max_parts,
+            first_part_no=first_part_no,
+        )
+
+        # Inspect PartNumber to verify results
+        assert [p["PartNumber"] for p in parts] == expected_parts, fail_msg
```
mephenor, line 522: So this is basically inline parametrization, the check only compares that the correct part numbers are returned and the fail_msg is there to figure out which test case regressed?

mephenor: `count` is mostly used as `count+1`, why not make that the constant and subtract once instead?

tests/integration/providers/objstorage/test_s3_objstorage.py
```diff
@@ -470,6 +470,56 @@ async def test_list_parts_invalid_args(s3: S3Fixture):
...
+    # Upload enough parts at minimum part size to exceed max pagination window size (1000)
+    for i in range(2, count + 1):  # Add 1 so `count` is accurate, not off by 1
+        await upload_part_of_size(
+            storage_dao=s3.storage,
+            upload_id=upload_id,
+            bucket_id=bucket_id,
+            object_id=object_id,
+            size=5 * 1024**2,
+            part_number=i,
+        )
```
mephenor, line 502: What's the runtime behavior here, i.e. time to populate and RAM usage?

tests/integration/providers/objstorage/test_s3_objstorage.py
```diff
@@ -470,6 +470,56 @@ async def test_list_parts_invalid_args(s3: S3Fixture):
...
+
+    # Perform tests inside a single setup instead of uploading all parts anew each time:
+    parts = []
+    for max_parts, first_part_no, expected_parts, fail_msg in [
+        (None, None, [n for n in range(1, count + 1)], "Basic pagination failed"),
+        (count, None, [n for n in range(1, count + 1)], "Failed when max_parts > 1000"),
+        (1, 1025, [1025], "Failed when targeting single part with PartNumber > 1001"),
```
mephenor: Shouldn't that be `>= 1001`?

mephenor (dismissed):

tests/integration/providers/objstorage/test_s3_objstorage.py
TheByronHimes: about a minute, maybe 5 gigs? We can also make the test conditional so it's only run manually

tests/integration/providers/objstorage/test_s3_objstorage.py
TheByronHimes: Yes, that is what's happening

tests/integration/providers/objstorage/test_s3_objstorage.py
TheByronHimes: I'm not sure if you're referring to the second or the third case, but either way no. The first full 1000-part window will return parts 1-1001, so the second window is touched only when starting with 1002 (first_part_no). Max_parts, by contrast, is 1-1000 and so the test is for some number over 1000.

mephenor: The third one.
With 0-based indexing we would have [0, 999] and with 1-based indexing the first 1000 parts should be [1, 1000], so 1001 would be in the next batch

tests/integration/providers/objstorage/test_s3_objstorage.py
mephenor: Let's do it that way.
It's a bit heavy to run each time

tests/integration/providers/objstorage/test_s3_objstorage.py
TheByronHimes: Oh you're right, I can't count.

mephenor (approved):

PR #230, opened 2026-05-22
Sort s3 mpu parts

TheByronHimes:
Updates the _list_parts() functionality so the list is returned sorted by part number. 

Bumps version to `8.2.1`


src/hexkit/providers/s3/provider/objstorage.py
```diff
@@ -431,7 +431,7 @@ async def _list_parts(
                 # If max_parts specified, return once amount is reached
                 if max_parts and len(parts) >= max_parts:
                     return parts[: max_parts + 1]
-            return parts
+            return sorted(parts or [], key=lambda p: p["PartNumber"])
```
mephenor: Why the `or []` here?
parts is already initialized as `parts: list[dict] = []`

mephenor (approved):

PR #231, opened 2026-06-10
DAO Pagination (GSI-2412)

TheByronHimes:
The DAO protocol and providers now expose pagination parameters in `find_all()`: `sort`, `skip`, and `limit`. Additionally, the result of `find_all()` exposes a `.total_count()` method that returns the total number of items matching the filter regardless of skip/limit. Together, these should enable both offset-based and cursor-based pagination.

Bumps the version to `8.3.0`

### DIFF
--- a/src/hexkit/protocols/dao.py
+++ b/src/hexkit/protocols/dao.py
@@ -223,7 +223,13 @@ class Dao(typing.Protocol[Dto]):
         """
         ...
 
-    def find_all(self, *, mapping: Mapping[str, Any]) -> AsyncIterator[Dto]:
+    def find_all(
+        self,
+        *,
+        mapping: Mapping[str, Any],
+        skip: int | None = None,
+        limit: int | None = None,
+    ) -> AsyncIterator[Dto]:
         """Find all resources that match the specified mapping.
 
         The values in the mapping are used to filter the resources, these are
@@ -235,10 +241,19 @@ class Dao(typing.Protocol[Dto]):
             mapping:
                 A mapping where the keys correspond to the names of resource fields
                 and the values correspond to the actual values of the resource fields.
+            skip:
+                Number of matching resources to skip before yielding results.
+                Defaults to None (no skipping).
+            limit:
+                Maximum number of resources to yield. Defaults to None (no limit).
 
         Returns:
             An AsyncIterator of hits. All hits are in the form of the respective DTO
             model.
+
+        Raises:
+            InvalidMappingError: If `mapping` doesn't pass validation.
+            ValueError: if `skip` or `limit` are less than 0.
         """
         ...
 
--- a/src/hexkit/protocols/dao.py
+++ b/src/hexkit/protocols/dao.py
@@ -20,7 +20,7 @@ with the database.
 
 import typing
 from abc import ABC, abstractmethod
-from collections.abc import AsyncIterator, Collection, Mapping
+from collections.abc import AsyncIterator, Awaitable, Callable, Collection, Mapping
 from contextlib import AbstractAsyncContextManager
 from functools import partial
 from typing import Any, Generic, Literal, TypeVar
@@ -35,6 +35,7 @@ __all__ = [
     "Dao",
     "DaoFactoryProtocol",
     "FindError",
+    "FindResult",
     "MultipleHitsFoundError",
     "ResourceAlreadyExistsError",
     "ResourceNotFoundError",
@@ -140,6 +141,39 @@ class DbTimeoutError(DaoError):
 UUID4Field = partial(Field, default_factory=uuid4)
 
 
+class FindResult(AsyncIterator[Dto]):
+    """An async-iterable result from find_all() that also exposes pagination metadata.
+
+    The `total_count()` method performs an additional count query on first call (lazily)
+    and caches the result for subsequent calls.
+    """
+
+    def __init__(
+        self,
+        *,
+        results_iterator: AsyncIterator[Dto],
+        get_total_count: Callable[[], Awaitable[int]],
+    ) -> None:
+        self._iterator = results_iterator
+        self._get_total_count = get_total_count
+        self._cached_total: int | None = None
+
+    def __aiter__(self) -> "FindResult[Dto]":  # noqa: D105
+        return self
+
+    async def __anext__(self) -> Dto:  # noqa: D105
+        return await self._iterator.__anext__()
+
+    async def total_count(self) -> int:
+        """Total number of matching documents, ignoring skip and limit.
+
+        The result is cached after the first call.
+        """
+        if self._cached_total is None:
+            self._cached_total = await self._get_total_count()
+        return self._cached_total
+
+
 class Dao(typing.Protocol[Dto]):
     """A duck type with methods common to all DAOs."""
 
@@ -233,7 +267,7 @@ class Dao(typing.Protocol[Dto]):
         skip: int | None = None,
         limit: int | None = None,
         sort: SortSpec | None = None,
-    ) -> AsyncIterator[Dto]:
+    ) -> "FindResult[Dto]":
         """Find all resources that match the specified mapping.
 
         The values in the mapping are used to filter the resources, these are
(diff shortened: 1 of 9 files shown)


src/hexkit/providers/mongodb/provider/dao.py
```diff
@@ -322,19 +331,47 @@ async def find_all(self, *, mapping: Mapping[str, Any]) -> AsyncIterator[Dto]:
...
+                or -1 (descending). Strongly recommended when using skip/limit.
+                Defaults to None (no sort).
 
         Returns:
-            An AsyncIterator of hits. All hits are in the form of the respective DTO
-            model.
+            A FindResult that is async-iterable and also provides total_count().
         """
+        skip = skip or 0
+        limit = limit or 0
```
Cito: maybe add a comment that limit=0 means unbounded in mongo (not all db engines do this)

src/hexkit/providers/mongokafka/provider/daopub.py
```diff
@@ -381,23 +390,58 @@ async def find_all(self, *, mapping: Mapping[str, Any]) -> AsyncIterator[Dto]:
...
+        dto_model = self._dto_model
+        collection = self._collection
+
+        async def _iter() -> AsyncIterator[Dto]:
+            with translate_pymongo_errors():
+                cursor = collection.find(filter=mapping)
+                if sort:
+                    cursor = cursor.sort(sort)
+                async for document in cursor.skip(skip).limit(limit):
+                    if document.get("__metadata__", {}).get("deleted", False):
```
Cito: here you filter after skipping and limiting - it should be the other way round

i guess the most performant way is to add a filter for (metadata missing or metadata.deleted false) upfront.

src/hexkit/providers/mongodb/provider/dao.py
```diff
@@ -322,19 +331,47 @@ async def find_all(self, *, mapping: Mapping[str, Any]) -> AsyncIterator[Dto]:
             mapping:
                 A mapping where the keys correspond to the names of resource fields
                 and the values correspond to the actual values of the resource fields.
+            skip:
+                Number of matching resources to skip before yielding results.
+                Defaults to None (no skipping).
+            limit:
+                Maximum number of resources to yield. Defaults to None (no limit).
```
Cito, line 337: The docstring should also define what limit=0 means - unbounded (as in mongo) or literally 0 (which I prefer since it#s much more intuitive). The implementation is currently for the former, the latter would need slight adjustment for mongo.

tests/unit/test_in_mem_dao.py
```diff
@@ -484,3 +484,179 @@ async def test_fields_with_dollar_sign():
...
+    results = [
+        x.title async for x in dao.find_all(mapping={}, skip=1, limit=2, sort=asc_sort)
+    ]
+    assert results == ["Banana", "Cherry"]
+
+    # skip larger than collection size returns no results
+    results = [x.title async for x in dao.find_all(mapping={}, skip=10, sort=asc_sort)]
+    assert results == []
+
+    # skip=0, limit=0 behaves like no pagination (limit=0 means no limit)
```
Cito: we should test both, limit=0 and limit=None and check that they do what we say in the docstring (for all providers)

tests/integration/providers/daopubsub/test_mongokafka_pubsub.py
```diff
@@ -963,3 +963,34 @@ async def test_unique_index_error_handling(mongo_kafka: MongoKafkaFixture):
 
         with pytest.raises(UniqueConstraintViolationError):
             await dao.upsert(dto5)
+
```
Cito, line 966: Here we should also add some pagination tests after adding raw documents with no metadata at all, and metadata.deleted = True

src/hexkit/protocols/dao.py
```diff
@@ -35,13 +35,17 @@
...
     "MultipleHitsFoundError",
     "ResourceAlreadyExistsError",
     "ResourceNotFoundError",
+    "SortSpec",
     "UUID4Field",
     "UniqueConstraintViolationError",
 ]
 
+SortSpec = list[tuple[str, Literal[1, -1]]]
+
```
Cito: We could use simple simple strings with a "-" prefix for descending order - that's how the Django ORM does it, so it's a well-established pattern in Python.

Cito (changes requested):
Thank you, that looks useful.

I have pointed out two edge cases that need some refinement, and made one suggestions regarding the sort spec.

src/hexkit/providers/mongodb/provider/dao.py
TheByronHimes: What is the use case for having a limit of literally 0?

src/hexkit/providers/mongokafka/provider/daopub.py
TheByronHimes: Ugh, yes, I don't know how I didn't catch that. Good spot

src/hexkit/providers/mongodb/provider/dao.py
Cito: Passing a value of 0 is not really useful (if you want unbound you can already just pass nothing/None). My main point is that the behavior should be well specified in the docstring. Options are 1) raise error, 2) do exactly what was requested (return nothing) or 3) re-interpret as "infinity". I think 2) is the most obvious one - Django ORM, Python slicing, Postgres and MySQL all work that way.

tests/integration/providers/daopubsub/test_mongokafka_pubsub.py
TheByronHimes: Yes, good call: [bccf1c2](https://github.com/ghga-de/hexkit/pull/231/commits/bccf1c27c6b6ff62fd4e434056686041e8a4c57f)

src/hexkit/protocols/dao.py
TheByronHimes: Yes, this is easier to type and to read. I have made the changes in [134691b](https://github.com/ghga-de/hexkit/pull/231/commits/134691bf06f6c665f1ad0ad23a7db9090ddd75f7)

src/hexkit/providers/mongodb/provider/dao.py
TheByronHimes: This should be solved by the clear doc string now.

tests/unit/test_in_mem_dao.py
TheByronHimes: There is now a test for that

src/hexkit/providers/mongodb/provider/dao.py
TheByronHimes: I've changed limit's behavior such that 0 returns no items and None means limitless. This is clearly documented. Changes in [491a25e](https://github.com/ghga-de/hexkit/pull/231/commits/491a25e992f78d300026ff8c599e10fe74e810dc)

tests/integration/providers/daopubsub/test_mongokafka_pubsub.py
```diff
@@ -963,3 +963,81 @@ async def test_unique_index_error_handling(mongo_kafka: MongoKafkaFixture):
...
+        docs = [ExampleDto() for _ in range(4)]
+        for dto in docs:
+            await dao.insert(dto)
+            await dao.delete(dto.id)
+
+        if delete_metadata:
+            db = mongo_kafka.mongodb.client.get_database(mongo_kafka.config.db_name)
+            collection = db["example"]
+            retrieved_docs = collection.find().to_list()
+            assert len(docs) == 4
```
Cito:
```suggestion
            assert len(retrieved_docs) == 4
```

tests/integration/providers/daopubsub/test_mongokafka_pubsub.py
```diff
@@ -963,3 +963,81 @@ async def test_unique_index_error_handling(mongo_kafka: MongoKafkaFixture):
...
+        # Run a query and make sure total_count() is 2
+        result = dao.find_all(mapping={})
+        page = [hit async for hit in result]
+        assert len(page) == 2
+
+        total = await result.total_count()
+        assert total == 2
+
+
+# @pytest.mark.parametrize("skip")
```
Cito:
```suggestion

```

Cito (dismissed):
All open issues were addressed. Found just one small mistake in the tests.

Cito (approved):

PR #233, opened 2026-06-26
Fix InMemDao sorting for nested fields (GSI-2454)

TheByronHimes:
Previously the InMemDao did a straight lookup by field name for sorting, which broke for nested fields. This fixes that.


Cito (approved):
I pushed a change to translate only the exact ID field to "_id" (not substring/nested matches), and sort missing values consistently instead of raising. This matches the real
MongoDB provider and removes the now unneeded string replacement. Also added tests for that.

You can merge after reviewing my changes.

PR #235, opened 2026-06-30
Fix outbox metadata handling (GSI-2474)

TheByronHimes:
Fixes a bug that prevented find_all in the outbox DAO provider from returning documents without metadata. Adds a regression test too.

### DIFF
--- a/src/hexkit/providers/mongokafka/provider/daopub.py
+++ b/src/hexkit/providers/mongokafka/provider/daopub.py
@@ -420,8 +420,20 @@ class MongoKafkaDaoPublisher(Generic[Dto]):
         validate_find_mapping(mapping, dto_model=self._dto_model)
         mapping = replace_id_field_in_find_mapping(mapping, self._id_field)
 
-        # Ensure we don't retrieve deleted docs
-        mapping_without_deleted = {**mapping, "__metadata__.deleted": False}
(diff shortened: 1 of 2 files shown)


src/hexkit/providers/mongokafka/provider/daopub.py
```diff
@@ -420,8 +420,20 @@ def find_all(  # noqa: C901
         validate_find_mapping(mapping, dto_model=self._dto_model)
         mapping = replace_id_field_in_find_mapping(mapping, self._id_field)
 
-        # Ensure we don't retrieve deleted docs
-        mapping_without_deleted = {**mapping, "__metadata__.deleted": False}
+        # Ensure we don't retrieve deleted docs. Documents lacking outbox metadata are
+        # treated as valid (matching update/delete), and the $and wrapper avoids
+        # clobbering any caller-supplied $or in the mapping.
```
Cito, line 425: Wouldn't a top-level `$or` key in `mapping` be rejected by `validate_find_mapping` anyway?

Cito (approved):
Thank you for fixing this so quickly.

src/hexkit/providers/mongokafka/provider/daopub.py
TheByronHimes: No, the validator ignores fields that start with `$`

PR #238, opened 2026-07-03
Move pagination into asyncio.to_thread

mephenor:
Currently the pagination and related IO accidentally happen synchronously on the main thread.
This PR moves the corresponding callsites completely into `asyncio.to_thread` which should correctly dispatch them on separate threads.

As a performance optimization, `_list_multipart_uploads_for_object` now uses a `Prefix=object_id` filter. This works for the current setup, but will break if we ever change *any* functionality that creates an object to use anything else than `Key=object_id`.


TheByronHimes (approved):
LGTM, but I think a template update is required now

PR #241, opened 2026-07-08
Reduce S3 Calls (GSI-2490)

TheByronHimes:
Re-evaluated every S3 provider operation and removed defensive existence checks where the main S3 call already fails with a translatable error.

Removed prechecks

- `delete_bucket` and `list_all_object_ids` rely on `NoSuchBucket`
- `list_parts`, `abort_multipart_upload`, and `complete_multipart_upload` rely on `NoSuchUpload`
- `get_object_etag`/`get_object_size`/`get_object_metadata` rely on `head_object` failing

Optimized

- `_assert_object_(not_)exists` now checks the object first and only checks the bus one call instead of two

Kept (with comments explaining why)

- `create_bucket` - S3 silently succeeds re-creating an owned bucket, so we let the app handle that error
- Presigned URL ops - signing is local, nothing server-side would raise if the upload/bucket didn't exist
- `init_multipart_upload` - to prevent duplicate uploads 
- `delete_object` - so the application decides how to interpret missing objects
- `copy_object` dest check - to prevent overwriting things by accident

### Note for near future

We should consider removing `MultipleActiveUploadsError`. The built-in check for multiple active uploads doesn't give us much. The methods already target specific uploads via `upload_id`, and if an application needs enforce upload exclusivity then that should be done application-side. That would mandate a major version bump though.

Version bumped to `8.5.0`

### DIFF
--- a/src/hexkit/providers/s3/provider/objstorage.py
+++ b/src/hexkit/providers/s3/provider/objstorage.py
@@ -549,8 +549,8 @@ class S3ObjectStorage(ObjectStorageProtocol):
                 + f" smaller or equal to {self.MAX_FILE_PART_NUMBER}"
             )
 
-        # Upload check needed: presigning is local-only, so a missing upload or
-        # multiple active uploads can only be detected here.
+        # Upload check needed: presigning is local-only, so a missing upload can
+        # only be detected here. Exclusiveness is not enforced: upload_id is explicit.
         await self._assert_multipart_upload_exists(
             upload_id=upload_id,
             bucket_id=bucket_id,
--- a/src/hexkit/providers/s3/provider/objstorage.py
+++ b/src/hexkit/providers/s3/provider/objstorage.py
@@ -859,14 +859,13 @@ class S3ObjectStorage(ObjectStorageProtocol):
         This only works reliably as long as there are no other ongoing multipart operations for
         the same destination bucket and object ID, in which case this should be set to false.
         """
-        file_size = await self._get_object_size(
-            bucket_id=source_bucket_id, object_id=source_object_id
-        )
-
-        # Destination object check needed: CopyObject silently overwrites; the size
-        # lookup above already validated the source.
+        # Before doing anything else, make sure the target doesn't already exist
         await self._assert_object_not_exists(
             bucket_id=dest_bucket_id, object_id=dest_object_id
+        )
+
+        file_size = await self._get_object_size(
+            bucket_id=source_bucket_id, object_id=source_object_id
         )
         part_size = calc_part_size(file_size=file_size)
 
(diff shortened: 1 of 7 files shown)


src/hexkit/providers/s3/provider/objstorage.py
mephenor, line 226: `content = await asyncio.to_thread(bucket.objects.all)` can be moved into the if block

src/hexkit/providers/s3/provider/objstorage.py
```diff
@@ -863,6 +867,8 @@ async def _copy_object(
             bucket_id=source_bucket_id, object_id=source_object_id
         )
 
+        # Destination object check needed: CopyObject silently overwrites; the size
+        # lookup above already validated the source.
         await self._assert_object_not_exists(
             bucket_id=dest_bucket_id, object_id=dest_object_id
         )
```
mephenor: Existence check can be moved in front of fetching the size.
It's meaningless to fetch it if we're not using it.

src/hexkit/providers/s3/provider/objstorage.py
```diff
@@ -907,6 +913,7 @@ async def _delete_object(self, *, bucket_id: str, object_id: str) -> None:
         """Delete an object with the specified id (`object_id`) in the bucket with the
         specified id (`bucket_id`).
         """
+        # Object check needed: DeleteObject reports success even for missing objects.
         await self._assert_object_exists(bucket_id=bucket_id, object_id=object_id)
```
mephenor: I'd argue the check can be removed:
We employ similar patterns across other repos and trying to delete an already deleted object should ideally be observably idempotent to just deleting the existing object.

It doesn't really make sense for this to fail.
Could also be reflected in the docstring

src/hexkit/providers/s3/provider/objstorage.py
```diff
@@ -830,17 +824,27 @@ async def _get_object_size(self, *, bucket_id: str, object_id: str) -> int:
     async def _get_object_metadata(
```
mephenor, line 824: `_get_object_etag` and `_get_object_size` have their corresponding public methods.
Maybe this one should also just have a public interface.
Code that needs to call both could save a roundtrip fetching the whole metadata and extracting locally, instead.

mephenor (changes requested):
Looks good, I just some additional places that could use slight improvement.

Do the existing tests not cover all the failure modes for the provider?
If so, should we extend them?

src/hexkit/providers/s3/provider/objstorage.py
TheByronHimes: I guess I agree. If the application needs to enforce object presence (the less common case) it can explicitly do that with the methods already available.

src/hexkit/providers/s3/provider/objstorage.py
TheByronHimes: Done

src/hexkit/providers/s3/provider/objstorage.py
TheByronHimes: Done

src/hexkit/providers/s3/provider/objstorage.py
TheByronHimes: Done

mephenor (approved):

PR #243, opened 2026-07-10
DB Version Check Helper (GSI-2456)

TheByronHimes:
This introduces a `check_db_version()` utility for the migration tools. Its purpose is to provide a convenient way to enforce DB state in application code independent of the checks that occur in migrations themselves.

### DIFF
--- a/src/hexkit/providers/mongodb/migrations/_utils.py
+++ b/src/hexkit/providers/mongodb/migrations/_utils.py
@@ -26,7 +26,11 @@ from pymongo.asynchronous.collection import AsyncCollection
 from pymongo.asynchronous.database import AsyncDatabase
 
 from hexkit.providers.mongodb.config import MongoDbConfig
-from hexkit.providers.mongodb.provider import document_to_dto, dto_to_document
+from hexkit.providers.mongodb.provider import (
+    ConfiguredMongoClient,
+    document_to_dto,
+    dto_to_document,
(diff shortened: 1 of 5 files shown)


src/hexkit/providers/mongodb/migrations/_utils.py
```diff
@@ -18,18 +18,107 @@
...
+    total_duration_ms: int
+
+
+class DbVersionMismatchError(RuntimeError):
+    """Raised when the database is not at the version expected by the service."""
+
+    def __init__(self, *, db_version: int, target_version: int):
+        msg = (
+            f"Database version is {db_version}, but the service expects version"
+            + f" {target_version}. Make sure migrations have been applied."
```
Cito: the "make sure migrations have been applied" is misleading when the database's version is too new instead of too old (the error appears in both cases)

Cito: i think that szenario is also untested in test_check_db_version_mismatch

Cito (dismissed):
Just one minor issue, see below.

src/hexkit/providers/mongodb/migrations/_utils.py
TheByronHimes: Fixed, thank you

Cito (approved):

PR #245, opened 2026-07-10
 Add Kafka test utils documentation

Cito:


user_guide/providers/kafka/test_utils.qmd
```diff
@@ -0,0 +1,400 @@
...
+(or compatible) daemon.
+
+Install hexkit with the `test-akafka` extra to use these utilities (it adds
+testcontainers and the pytest plugins on top of the `akafka` extra):
+
+```bash
+pip install hexkit[test-akafka]
+```
+
+For unit tests that should not touch a broker at all, hexkit also ships an
```
TheByronHimes:
```suggestion
For unit tests that should not use a live broker, hexkit also ships an
```

user_guide/providers/kafka/test_utils.qmd
```diff
@@ -0,0 +1,400 @@
...
+validation that `expect_events()` runs on exit.
+
+### Managing Broker State
+
+- `clear_topics(topics=..., exclude_internal=True)` deletes all events from
+  the given topic(s), or from all topics when none are specified (internal
+  `__`-prefixed topics are excluded by default). The `kafka_fixture` calls
+  this before every test; call it yourself to separate phases within one test
+  or when using the persistent fixture. Compacted topics are handled
+  transparently by temporarily switching their cleanup policy.
```
TheByronHimes:
```suggestion
  by temporarily switching their cleanup policy.
```

user_guide/providers/kafka/test_utils.qmd
```diff
@@ -0,0 +1,400 @@
...
+        ):
+            await emitter.greet(user_id="abc123", name="Jane")
+```
+
+### Testing the Consuming Side
+
+Publish an inbound event with `publish_event()`, then run a real
+[`KafkaEventSubscriber`](subscriber.qmd) with `run(forever=False)`, which
+consumes exactly one event and returns. Assert on the side effects in the
+service's core:
```
TheByronHimes:
```suggestion
Publish an inbound event with `publish_event()`, then run a real
[`KafkaEventSubscriber`](subscriber.qmd) with `run(forever=False)`, which
consumes exactly one event and returns. Note that because the consumer waits for an event even if `forever=False`, your tests might hang if no event is published to the specified topic. Assert on the side effects in the service's core:
```
This is a detail worth including and something we should patch at some point.

TheByronHimes (changes requested):

TheByronHimes (approved):

PR #249, opened 2026-07-21
Otel changes and testutils

mephenor:
Changed OpenTelemetry config code to manually instrument all installed instrumentation libraries.
This should no longer be used in conjunction with opentelemetry-instrument, which was fragile, either way.

Also removed explicitly setting an env variable to disable OTel as this was no longer working with current OTel versions, as the code evaluating the env variable now runs before our code during autoinstrumentation.
Instead simply skipping `TracerProvider` creation should serve the same purpose.

Added testutils (and some sanity tests for those) that can be used in the services to build tests around span capture.

Not really sure we need both of the variants, i.e. TracerProvider already exists vs. create a new one.
Might make sense in a unit vs. integration test scenario, where the integration tests hit the normal startup code.

Changes should be effectively non-breaking, so this only bumps the minor version.


src/hexkit/opentelemetry/testutils.py
```diff
@@ -0,0 +1,140 @@
...
+    yield OpenTelemetryFixture(exporter=exporter)
+
+
+def get_otel_provider_fixture(
+    scope: PytestScope = "session", name: str = "otel_provider"
+):
+    """Get a fixture with desired scope and name that sets up span capturing.
+
+    By default, the session scope is used since only one `TracerProvider` can be
+    configured per process.<
```
ckaipf:
```suggestion
    configured per process.
```

src/hexkit/opentelemetry/testutils.py
```diff
@@ -0,0 +1,140 @@
...
+    if not isinstance(provider, TracerProvider):
+        # No app under test has configured OpenTelemetry yet, so do it here, wiring
+        # up the in-memory exporter directly instead of the default OTLP exporter -
+        # this avoids pointless connection attempts against a collector in tests.
+        configure_opentelemetry(
+            service_name=service_name,
+            config=OpenTelemetryConfig(enable_opentelemetry=True),
+            span_processor=SimpleSpanProcessor(exporter),
+        )
+    else:
```
ckaipf, line 99: The else branch of _otel_provider_fixture adds a SimpleSpanProcessor to an existing provider and never removes it on teardown (there's no code after yield). Fine for session scope, but worth a comment that it's intentionally not cleaned up.

ckaipf:
`getattr(module, class_name) `and `.instrument()` in the loop aren't guarded. If a pinned instrumentor renames its class in a future bump, getattr raises AttributeError and aborts startup mid-loop. Given the pins it's low-risk, but you may want one instrumentor's failure not to take down the whole loop (catch and log per-entry).

ckaipf:
FastAPI (and other class-patching) instrumentation is now order-sensitive. This is the one real behavioral trap. FastAPIInstrumentor().instrument() patches fastapi.FastAPI so only apps instantiated after the call get traced. Under the old opentelemetry-instrument wrapper, patching happened at interpreter startup before any user code, so ordering was guaranteed. Now configure_opentelemetry() must run before the service constructs its FastAPI app, or requests won't be traced.

mephenor:
> FastAPI (and other class-patching) instrumentation is now order-sensitive. This is the one real behavioral trap. FastAPIInstrumentor().instrument() patches fastapi.FastAPI so only apps instantiated after the call get traced. Under the old opentelemetry-instrument wrapper, patching happened at interpreter startup before any user code, so ordering was guaranteed. Now configure_opentelemetry() must run before the service constructs its FastAPI app, or requests won't be traced.

I don't think we can avoid that.
Autoinstrumentation via `opentelemetry-instrument` happens at import time, while everything we have now happens at runtime.
Might be worth a comment, though.

Edit:
See 6aeb4a9811169e31c5ec999ab8bc8d112c98668f

mephenor:
> `getattr(module, class_name) `and `.instrument()` in the loop aren't guarded. If a pinned instrumentor renames its class in a future bump, getattr raises AttributeError and aborts startup mid-loop. Given the pins it's low-risk, but you may want one instrumentor's failure not to take down the whole loop (catch and log per-entry).

Addressed in dc01a389835d9bf0db0308a81f5e8dd557129ed5.

src/hexkit/opentelemetry/testutils.py
mephenor: This is only populated for the tests and there's no stable public API to tear it down again.
I'd leave it as is for now.

ckaipf (approved):

PR #251, opened 2026-07-31
Add observability tools docs

Cito:


user_guide/observability_tools/opentelemetry.qmd
```diff
@@ -0,0 +1,174 @@
...
+configure_opentelemetry(service_name="my-cool-special-service", config=config)
+```
+
+If `enable_opentelemetry` is `False`, this is a no-op and the OpenTelemetry API's
+default no-op tracer remains in place — all instrumentation calls in hexkit and your
+service become inert, with negligible overhead. If enabled, it sets up a
+`TracerProvider` carrying the service name as its resource attribute, installs the
+sampler and exporter described below, and finally auto-instruments every supported
+library that is installed (matching the extras above) — instrumentation libraries
+that are not installed are silently skipped.
```
mephenor: Not fully true, there's a debug log message for every instrumentation not found in  the local context and if instrumenting a library fails, a warning is logged.

user_guide/observability_tools/opentelemetry.qmd
```diff
@@ -0,0 +1,174 @@
...
+
+::: {.callout-warning}
+`configure_opentelemetry` must run **before** any objects to be instrumented are
+constructed. In particular, a FastAPI app created before this call will silently
+miss its instrumentation.
+:::
```
mephenor, line 104: I'd drop the second sentence here, as it's not clear when order can be relevant without inspecting the instrumentation code and this special cases FastAPI, while it also applies to other instrumentation libraries.

In general, auto-instrumentations that patch a class need the instrumentation to run first, else an uninstrumented object is created.

user_guide/observability_tools/logging.qmd
```diff
@@ -19,4 +19,154 @@ title: "Configurable Logging"
...
+The module is part of hexkit's core, so no extra is needed. Services keep using the
+standard library as usual, i.e. `logging.getLogger(__name__)` and the familiar logging
+calls, and only call hexkit once at startup to configure the output.
+
+## Configuration
+
+All settings are collected in
+[`LoggingConfig`](../../reference/log.LoggingConfig.qmd), a
+`pydantic_settings.BaseSettings` class. Inherit your service's config class from it,
+or instantiate it directly:
```
mephenor: Is the `pydantic_settings.BaseSettings` relevant here to the reader?

user_guide/observability_tools/logging.qmd
```diff
@@ -19,4 +19,154 @@ title: "Configurable Logging"
...
+config = LoggingConfig(
+    log_level="INFO",
+    service_name="my-cool-special-service",
+    service_instance_id="germany-bw-instance-001",
+)
+```
+
+| Parameter | Description |
+| --------- | ----------- |
+| `log_level` | The minimum log level to capture: `CRITICAL`, `ERROR`, `WARNING`, `INFO` (the default), `DEBUG`, or `TRACE`. |
```
mephenor, line 60: Are we even using `TRACE` anywhere/what would trace be used for?

user_guide/observability_tools/logging.qmd
```diff
@@ -19,4 +19,154 @@ title: "Configurable Logging"
...
+| `service_name` | The name of the (micro-)service, included in every log message. Required. |
+| `service_instance_id` | A string uniquely identifying this instance among all instances of the service, included in every log message. Required. |
```
mephenor, line 62: Breaking abstractions a bit here, because I already got confused by this:

In our deployment this would be shared by instances in a ReplicaSet, if I'm not mistaken.
Though probably not the right place to document/discuss this.

user_guide/observability_tools/logging.qmd
```diff
@@ -19,4 +19,154 @@ title: "Configurable Logging"
...
+| `log_format` | If set, replaces the JSON output with a classic [format string](https://docs.python.org/3/library/logging.html#logrecord-attributes) (see [Plain-Text Output](#plain-text-output)). Defaults to `None`. |
+| `log_traceback` | Whether to include exception tracebacks in log messages. Defaults to `True`. |
+
+`TRACE` is a custom level that hexkit registers below `DEBUG` (numeric value 5).
+Since importing `hexkit.log` registers the level name, `getLevelName("TRACE")` and
+level-based filtering work with it like with the built-in levels.
```
mephenor: See above

user_guide/observability_tools/logging.qmd
```diff
@@ -19,4 +19,154 @@ title: "Configurable Logging"
...
+level and format. By default the root logger is configured, so records from all
+loggers in the service (including those of libraries) are captured; pass an
+explicit `logger` argument to configure only a specific logger instead.
+
+Right after setup, one `INFO` message ("Logging configured, complete configuration
+in details") is emitted with the full dump of the passed config object in its
+`details`. Since services typically pass their complete service config (which
+inherits from `LoggingConfig`), this documents the exact configuration each
+instance is running with. Values of Pydantic's `Secret*` types are masked in this
+dump, so credentials do not leak into the logs.
```
mephenor: Did we double check if everything that should be a `Secret*` type in this context actually is one?

mephenor (changes requested):

user_guide/observability_tools/opentelemetry.qmd
Cito: Good point. Fixed in 8cd3ddd.

Small correction on the second half: the WARNING only fires when an installed module doesn't expose the instrumentor class we expect. It doesn't cover instrumentation failing. `instrumentor_class().instrument()` sits outside the try, so an instrumentor that raises propagates out of configure_opentelemetry.

user_guide/observability_tools/opentelemetry.qmd
Cito: Actually it doesn't seem to apply in general: pymongo, httpx and aiokafka all patch at class/module level, so objects built before configure_opentelemetry still get traced. FastAPI's the odd one out: it swaps the FastAPI class itself, so only apps built after the call are instrumented. So it's really the only case where this bites. I kept it but explained why in f1bb3ed.

user_guide/observability_tools/opentelemetry.qmd
mephenor: Sorry, mixed up the details when looking at the code.
Looks good now.

user_guide/observability_tools/opentelemetry.qmd
mephenor: I was meaning to say there are such classes and that's why in general one should take care of running the instrumentation before instantiation.

I like the rephrased version, doesn't single out FastAPI.

user_guide/observability_tools/logging.qmd
Cito: You're right, it doesn't tell the reader anything on its own. We had a proper explanation already, just buried in the Kafka chapter. I've pulled it out into a general Configuration chapter (5b1edbb) that covers env vars, `config_from_yaml` and the source precedence, and the config pages now just link there instead of naming the base class.

user_guide/observability_tools/logging.qmd
Cito: Nothing in hexkit logs at `TRACE`. It's only registered as a level name, so services can use a tier below `DEBUG` if they want one. Added a note about that in b5933b8.

user_guide/observability_tools/logging.qmd
Cito: The docs just restate what the field asks for, which is a value unique per instance. It's not only a log label either, but goes into the Kafka client ID together with `service_name`, so sharing it across pods makes the broker-side client metrics ambiguous as well. If our deployment sets the same value for a whole ReplicaSet, that seems like something to fix in the chart rather than in hexkit, which shouldn't know anything about how we deploy it.

user_guide/observability_tools/logging.qmd
Cito: Checked all the config classes. Everything with an actual credential is already a `Secret` type (`kafka_ssl_password`, `s3_secret_access_key`, `s3_session_token`, the two vault ones, plus `mongo_dsn` and `redis_url` as `Secret[...]` since the DSNs can carry credentials). The remaining plain strings are just paths, endpoints or identifiers. `s3_access_key_id` is the only debatable one, but that's an identifier that goes over the wire in the clear anyway, so I left it.

Also ran it to be sure rather than trusting the annotations, config with Kafka/Mongo/S3/Vault, `configure_logging`, nothing leaked. Reworded the docs in 5b1edbb so the promise is "`Secret`-typed fields are masked" instead of a blanket "credentials don't leak".

mephenor: Nice, thank you for your diligence.

I didn't realize this prints the whole config, so we should keep that in mind when setting up new services that hold confident information inside their config.

Or we change the behavior in hexkit down the line, so it only prints the logging config part, depending on what the actual intent was here.

user_guide/observability_tools/logging.qmd
mephenor: Just something to keep in mind.

Might make sense to sync up with the ops part of the team to check how this is actually used, but I couldn't use it for the EMTS cleanup logic for this reason.

user_guide/configuration.qmd
```diff
@@ -0,0 +1,158 @@
...
+| -------- | ------ |
+| 1 | Keyword arguments passed to the constructor. |
+| 2 | Environment variables, prefixed as described below. |
+| 3 | Secret files in a dotenv file, `.env` in the current directory or `/secrets/.env` (configurable via the decorator's `dotenv_prefix`). |
+| 4 | A config YAML at the path given by the `{PREFIX}_CONFIG_YAML` environment variable. |
+| 5 | A config YAML named `.{prefix}.yaml` in the current directory, else in the home directory. |
+| 6 | The defaults declared on the config classes. |
+
+A path passed as `Config(config_yaml=...)` replaces the lookup in rows 4 and 5, and
+raises `ConfigYamlDoesNotExist` if no file is there.
```
mephenor:
```suggestion
raises `ConfigYamlDoesNotExist` if no file exists at that given path.
```

user_guide/configuration.qmd
```diff
@@ -0,0 +1,158 @@
...
+
+### The Prefix
+
+The `prefix` argument does double duty: it names the config YAML (`.my_service.yaml`,
+`MY_SERVICE_CONFIG_YAML`) *and* it becomes the environment variable prefix. Once a
+config class is decorated, the unprefixed variables shown above no longer apply —
+the settings are read from `MY_SERVICE_SERVICE_NAME`, `MY_SERVICE_KAFKA_SERVERS`,
+and so on. Nested values use a double underscore as separator, e.g.
+`MY_SERVICE_SOME_OBJECT__SOME_FIELD`. The prefix defaults to `ghga_services`.
```
mephenor: Could make it slightly more obvious, that the example prefix here is `MY_SERVICE/my_service` for first time users and add a short paragraph addressing case (in)sensitivity when a) specifying the prefix and b) using it in an env variable/config file name, i.e. what needs to be strictly upper case and what strictly lower case.

That lead to some initial confusion in the past, so might be something someone new to hexkit might stumble upon.

Cito: Good point, especially since all three ways to get this wrong fail silently. The value just isn't picked up and you get the default. Tested each one: setting variables are case-insensitive (`MY_SERVICE_KAFKA_SERVERS` and `my_service_kafka_servers` both work), but `MY_SERVICE_CONFIG_YAML` has to be exactly upper case because it's read straight from `os.environ` rather than through pydantic, and the YAML file name uses the prefix verbatim, so it's `.my_service.yaml`. Added a table covering all of that in 1e97be3f, and made it clear the `my_service`/`MY_SERVICE` in that section come from the `prefix=` in the example rather than being fixed strings.

mephenor (approved):

